PR #7, opened 2026-07-04
GPU box: count-toggled g6.xlarge, DeepSeek smoke tooling, repetition-collapse guard

ymohanty:
## Summary
- **Terraform** (`deploy/terraform/gpu.tf`): on-demand GPU box for DeepSeek OCR batch runs and demo windows. `gpu_box_count = 0/1` toggles it — default 0, so a plain `apply` never starts GPU billing (~$1/hr). Built from the Deep Learning **Base** AMI (Ubuntu 24.04): NVIDIA driver, Docker, and nvidia-container-toolkit preinstalled. SSH-only security group, ephemeral public IP (create/destroy workflow, nothing stateful), instance profile reusing the CPU-box IAM role. Pre-flight verified: AMI pattern resolves in ap-south-1, G-instance vCPU quota is 768, and `terraform plan` against live state shows **2 to add, 0 to change, 0 to destroy**.
- **Repetition-collapse guard** (`janasunani/pipeline/ocr_quality.py`): adopts the DSI report's DeepSeek signature-failure rule, generalized — their top-trigram share > 0.5 only catches single-word loops (a k-word phrase loop plateaus at ~1/k), so the detector keys on the **repeated-trigram share** (~1.0 for loops of any phrase length, ~0 for prose) at the same 0.5 threshold, with a 20-word floor. Collapsed pages store no text and are recorded in `unreadable_pages` so resumes skip them. Pure stdlib at the light `janasunani.pipeline` level, so CI imports it without extras.
- **Smoke tooling** (`scripts/gpu_smoke.sh` + README runbook): runs the 2-file DVC sample through format classification in the `pipeline-core` env, then DeepSeek OCR in the `ocr-deepseek` env, against one SQLite artifact — the deploy pattern the `[tool.uv].conflicts` split forces. Exits nonzero unless both pages extract.

## Not in this PR
The first real `apply` + on-box smoke run (starts billing; runbook in `deploy/terraform/README.md`). Known risk to watch there: DeepSeek's `trust_remote_code` may import `flash_attn` unconditionally even with eager attention forced.

## Test plan
- [x] 50 pytest green locally (5 new: prose, single-word collapse, phrase-loop collapse, short-text floor, degenerate inputs)
- [x] ruff clean
- [x] `terraform fmt -check` + `validate` + `plan` (2 add / 0 change / 0 destroy)
- [ ] On-box: `bash scripts/gpu_smoke.sh` after first apply

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ymohanty:
Codex finding valid, fixed in ce9bbdf: the smoke now requires every discovered page to extract (ok-count == row-count, non-empty), not just one.

PR #10, opened 2026-07-06
Script: sample English-subject complaints + documents into a zip

ymohanty:
## Summary
- `scripts/sample_english_complaints.py` — samples N complaints (default 10) whose **grievance subject is English**, via three gates: no Odia codepoints → ≥2 common English stopwords (rejects romanized Odia even when langdetect guesses 'en') → langdetect `en` (seeded, deterministic). Only complaints with ≥1 **STANDARD-storage-class** S3 document qualify (GLACIER keys are skipped, per the known bucket state). Output: one zip with `documents/<s3-key>` (key paths preserved so nested tickets keep the directory structure the pipeline's ticket parsing needs) + `complaints.parquet` (full metadata rows).
- Reads the Parquet lake locally; S3 via the existing `S3Service` (default credential chain). langdetect isn't a base dep — run with `uv run --with langdetect python scripts/sample_english_complaints.py`.

## Verification
- [x] Real run against the lake + bucket: 10/10 picked subjects verifiably English (188 shuffled candidates checked), 10 documents downloaded, 0.8 MB zip with the parquet inside
- [x] 5 new tests for the language gate (English accepted; Odia script, mixed script, romanized Odia, too-short rejected) — full suite 68 passed
- [x] ruff clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,13 +95,15 @@ uv run --extra pipeline-core janasunani-evaluate-pii --gold <gold.jsonl> # gate:
 ### 5 · Sample English complaints + documents (evaluation bundles)
 
 ```bash
-uv run --with langdetect python scripts/sample_english_complaints.py   # --n/--seed/--out
+uv run --extra pipeline-core python scripts/sample_english_complaints.py   # --n/--seed/--out
 ```
 
-Picks N complaints whose grievance subject is English (not Odia, not romanized
-Odia), downloads their S3 documents (STANDARD storage class only), and writes
-one zip: the documents + a `complaints.parquet` of their metadata. Selection
-logic and gates: [scripts/README.md](scripts/README.md).
+Picks N complaints that are English on both sides — the grievance subject (not
+Odia, not romanized Odia) *and* the scanned document (judged by the pipeline's
+format classifier + page-type ViT; documents that are pure PII like an Aadhaar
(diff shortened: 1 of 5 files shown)


ymohanty:
Updated per review: the sample is now English on **both sides** — after the subject gate, each document is judged by the pipeline's own models (format-classifier language must be exactly English on a majority of pages; page-type ViT must find ≥1 signal-class page, which drops pure-PII docs like Aadhaar/voter-ID scans). Gate evidence ships in the zip's parquet. Real run: 10/10 qualifying picks from 331 candidates, 11 docs dropped with logged reasons. Also fixes a real macOS hang the combination surfaced (xgboost+torch dual-OpenMP — OMP_NUM_THREADS=1 on darwin).

ymohanty:
Follow-up after user spot-check found Odia-dominant pages in the bundle: page language is now judged by the raw tesseract eng-vs-ori dominance signal (`perform_ocr`), not the format classifier's ~76%-accuracy label (verified lying: a doc it scored 1.00-English is 2/4 Odia pages), and the rule is strict — any Odia-dominant page rejects the document (`Sparse` stamp/signature pages tolerated; ≥1 confidently-English page required). Real rerun: 10/10 clean picks; drop log catches the earlier offenders; the PII-only gate fired on a real all-Identification doc. 73 tests green.

PR #27, opened 2026-07-10
Demo integration: live frontend + Makefile targets + inference docs

ymohanty:
Bundles the demo-integration wiring onto one branch (absorbs the Makefile work from the now-closed #28).

## Frontend → live API
- Next.js UI wired to `janasunani-api-live` against the frozen `serving/schemas.py` contract; stale mock copy removed; `ResultView` notes the `method:"fallback"` routing case.
- `.env.local.example` + README document `NEXT_PUBLIC_API_URL`.

## Makefile — live-demo targets
- `make models` (scoped DVC pull), `preflight`, `db`, `api`, `frontend`, **`up`**, `down`.
- **`make up`**: serve API (background) + frontend (foreground), one `Ctrl-C` traps and reaps both.
- **`make down`**: also stops the frontend (kill by `FRONTEND_PORT` — `pkill` can't match Next's forked `next-server`).
- **`db` is idempotent** (create-if-missing / start-if-stopped / re-migrate) and a prerequisite of `api`/`up`, so neither serves against a nonexistent database. An off-box `OLTP_DB_URL` skips local provisioning and is **never auto-migrated** (protects prod).
- Collapsed `OLTP_URL`/`OLTP_DB_URL` to one canonical `OLTP_DB_URL` that `?=` honors from env/`.env`, so preflight/db/api can't disagree — addresses **Codex #28 P2**.

## Docs
- New `janasunani/inference/README.md` (warm processor, fail-closed `build_processor`/`preflight`, error mapping, store selection).
- README module index links inference; serving README cross-links it.
- `docs/DEMO.md` gains a Makefile fast-path callout; frontend README notes `make up`.
- ROADMAP Phase 12 ⬜ → 🔄.

Base: `feat/demo-integration`. Supersedes #28.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #28, opened 2026-07-10
build(make): live-demo targets wrapping the DEMO.md runbook

ymohanty:
Codifies the `docs/DEMO.md` bring-up as Makefile targets (previously only raw `uv run --extra demo …` commands). Targets into `feat/demo-integration`.

## Targets (prefix-free — the whole group is the demo stack)
- `make models` — scoped DVC pull of **only** the model `.dvc` targets (never the PII-bearing raw data, unlike the unqualified `make pull`)
- `make preflight` — fast readiness check
- `make db` — throwaway Postgres (not the prod volume, not the :5433 test fixture) + `alembic upgrade head`
- `make api` — live real-inference API (prereq: `preflight`)
- `make frontend` — Next.js UI against the live API
- `make down` — tear down API + throwaway DB

## Notes
- **`.PHONY` is required**: `models` and `frontend` share names with repo directories, so without it make reports "up to date" and skips the recipe.
- **api/frontend stay discrete foreground targets** — make isn't a process supervisor; the one-command compose orchestration belongs to the deferred CPU-box work.
- Thin wrappers over already-tested console scripts (no new test surface).

Verified: `make -n` on each (PHONY defeats both dir collisions), help block renders, and a real `make preflight` passes against the local mirror.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,13 @@ RAW_LOCAL      ?= data/raw/
 EXHIBITS_LOCAL ?= outputs/
 INCOMING_REMOTE ?= $(BOX_REMOTE):'$(BOX_PROJECT_ROOT)/Data/Raw/'
 EXHIBITS_REMOTE ?= $(BOX_REMOTE):'$(BOX_PROJECT_ROOT)/Analysis/Exhibits/'
+# Live-demo stack (see docs/DEMO.md). Throwaway Postgres only — never the prod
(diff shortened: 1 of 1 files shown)


ymohanty:
Superseded by #27: the Makefile live-demo targets were merged into feat/demo-frontend-live, where Codex's P2 (honor OLTP_DB_URL across preflight/db/api) is also being addressed. Closing so the Makefile has a single home.

PR #39, opened 2026-07-22
fix(setup): preflight-check poppler + tesseract, fail early with install command

ymohanty:
## Problem

A collaborator ran `make setup` successfully, then hit this at pipeline runtime:

```
pdf2image.exceptions.PDFInfoNotInstalledError: Unable to get page count. Is poppler installed and in PATH?
```

`scripts/setup.sh` installs `uv`/`rclone`/`aws` into `~/.local`, but never handled the **poppler** (`pdfinfo`/`pdftoppm`) and **tesseract** system binaries the OCR pipeline shells out to. Those were only documented as a manual step in `docs/DEMO.md`, so setup reported success and the failure surfaced later on the first PDF.

## Fix

Add `check_system_ocr_deps()`, called from `install_missing_prereqs`:

- **Debian/Ubuntu/WSL (apt-get present):** installs `poppler-utils tesseract-ocr tesseract-ocr-ori` directly via `sudo apt-get`. `sudo` elevates only that command — consistent with the existing `sudo apt-get install git` guidance and not in conflict with the "don't run setup as root" guard (which protects the `~/.local` installs).
- **No apt (macOS / other distros):** prints the platform-appropriate install command (`brew` on macOS) and exits, rather than crashing later at runtime.
- **Install failure:** falls back to manual instructions + exit.
- Non-fatal warning (`warn_missing_odia_traineddata`) when the tesseract `ori` (Odia) traineddata is missing, since eng-only OCR still runs.

## Verification

- `bash -n scripts/setup.sh` passes.
- Exercised both branches with `apt-get`/`sudo` stubbed: apt path runs `apt-get update` + `install -y poppler-utils tesseract-ocr tesseract-ocr-ori`; no-apt path prints manual instructions and exits 1.

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -112,33 +112,46 @@ install_aws() {
 }
 
 # Poppler (pdfinfo/pdftoppm) and tesseract are system packages the OCR pipeline
-# shells out to at runtime. Unlike uv/rclone/aws they can't be installed into
-# ~/.local without sudo, so setup doesn't install them — it fails early with the
-# platform's install command rather than letting the pipeline crash on the first
-# PDF (pdf2image PDFInfoNotInstalledError). See docs/DEMO.md.
+# shells out to at runtime; without them it crashes on the first PDF (pdf2image
+# PDFInfoNotInstalledError). Unlike uv/rclone/aws they need a system package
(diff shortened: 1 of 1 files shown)


ymohanty:
@codex review

PR #29, opened 2026-07-11
Automate CI -> GHCR -> box deploy of the demo stack

ymohanty:
## Summary
- Frontend/API Docker images (`frontend/Dockerfile`, `deploy/api.Dockerfile`) — standalone Next.js output, `uv sync --extra demo` with BuildKit SSH for the private `dpic` dependency, models/data kept as host bind-mounts (never baked in).
- `deploy/docker-compose.yml` grows to `oltp` + `api` + `frontend` + `proxy` (Caddy: nip.io TLS via automatic Let's Encrypt, site-wide `basic_auth` since production grievance data must not be openly public). `oltp` service/volume left byte-identical to the Week-1 bring-up.
- `deploy/deploy.sh` — the only sanctioned box-side up-path: pulls, brings the stack up, blocks until `/health` reports the real pipeline processor, never `docker compose down`.
- `deploy/terraform/ci.tf` — GitHub OIDC provider + `janasunani-ci-deploy` IAM role, scoped to opening/closing port 22 on the CPU box's security group only, for the duration of a deploy job.
- `.github/workflows/deploy.yml` — `workflow_dispatch`-only (never runs on PR/push), optional `image_tag` input for a build-free redeploy/rollback.
- `tests/test_deploy_stack.py` — real-code-path checks on the actual compose file, Caddyfile, and shell scripts on disk.
- Docs: `docs/DEPLOY.md` §4 "Automated demo deploy" (architecture, one-time box setup, secrets/vars table, routine flow, rollback, known gaps), plus `deploy/README.md`, `deploy/terraform/README.md`, `docs/DEMO.md`, `docs/ROADMAP.md` updated to match.

## Test plan
- [x] `uv run pytest` — 242 passed, 1 skipped (Postgres-path test skips without a local throwaway DB; expected)
- [x] `uv run ruff check .` — all checks passed
- [x] `cd frontend && npm run lint && npm run build` — clean; standalone `server.js` emitted, `/api` confirmed baked into the client bundle (not a stray `127.0.0.1:8000`)
- [x] `docker build -f frontend/Dockerfile frontend` — real build, ran the resulting container, confirmed HTTP 200 and the baked `/api` in `.next/static`
- [x] `docker buildx build --check -f deploy/api.Dockerfile .` — Dockerfile lint, no warnings (hadolint not installed locally)
- [x] `terraform -chdir=deploy/terraform fmt -check && terraform -chdir=deploy/terraform validate` — clean
[...]

PR #6, opened 2026-07-04
Rebuild PII redaction on Presidio + evaluation gate

ymohanty:
## Summary
- Replaces the lost DSI Transformer-CRF PII stage (weights + training data were Box-only, unrecoverable) with a Presidio-based stage keeping the same contract: `pages.extracted_text` → `pages.redacted_text`.
- Custom recognizers for Indian grievances: mobile (`IN_MOBILE`), Aadhaar (`IN_AADHAAR`), PAN (`IN_PAN`), plus Presidio's email/phone and spaCy NER for person names. Typed tokens (`[NAME]`, `[PHONE]`, `[AADHAAR]`, `[PAN]`, `[EMAIL]`) preserve sentence structure for downstream summarize/categorize.
- Improvements over legacy: no 512-token truncation window, mixed "English, Odia" pages covered, explainable hits, bounded SQL batching. Everything in-process — citizen text never leaves the box.
- Adds a PII evaluation gate (`janasunani-evaluate-pii`): scores gold-labeled JSONL against the **production analyzer** (`detect_pii_spans` shares the same engine as `redact_text`), with the legacy DSI baseline (80.56% any-overlap coverage) as the pass/fail bar.
- Deletes the legacy `models/pii_tagger` mirror; `dvc.yaml` pipeline-sample stage now includes redaction.
- ROADMAP updated: Presidio rebuild status, DSI technical-report baselines, pre-backfill eval TODO.

## Test plan
- [x] 55 pytest green locally (incl. 4 redaction tests: Indian patterns, no-truncation, mixed-language, idempotency; 6 new eval/config tests)
- [x] ruff clean
- [x] `janasunani-evaluate-pii --help` functional
- [ ] Label ~100–200 real pages and run the gate vs the 80.56% baseline (queued, pre-backfill)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4, opened 2026-07-03
Phase 5: document-processing pipeline refold, OLTP exporter, model salvage

ymohanty:
## Summary

Phase 5 of the roadmap: the DSI document-processing pipeline lives in this repo, runs end-to-end on real documents (CPU stages), and its outputs flow into the OLTP store and the Parquet lake. Also contains the DSI-loss contingency (team disbanded; Box gone).

**Pipeline refold (`janasunani/pipeline/`)**
- `document_pipeline` copied verbatim from the DSI repo's `pipeline` branch, imports renamed; CLI (`janasunani-pipeline`), `PipelineConfig`, `STAGE_ORDER`, lazy per-stage imports, and the internal SQLite artifact DB all preserved.
- Dep groups populated with hard pins; `ocr-deepseek` (transformers 4.46.3) declared **mutually exclusive** with `pipeline-core`/`categorizer` (>=4.57,<5) via `[tool.uv].conflicts` — one Docker image per extra on deploy. scikit-learn pinned to the pickled-estimator era.
- **Verified on real grievance documents from S3** (format classifier + pytesseract): 5/5 pages extracted, including Odia (0 → 2,477 chars after adding `ori` traineddata + a warn-and-fall-back fix for missing tesseract languages — previously silent empty text marked ok).

**OLTP exporter + lake**
- `pages`/`documents` ORM models + Alembic revision `dd0eea37cac6` (index rule: keys are short identifiers only; unbounded text never enters a btree key). Schema applied on the cloud box.
- `janasunani-export-pipeline`: portable ON CONFLICT DO UPDATE upsert; re-export refreshes rows in place. `materialize` carries the new tables into the lake.
- `pipeline-sample` DVC stage: the CPU core stages over 2 DVC-tracked real documents → a versioned SQLite artifact; a living regression check on real data.

**DSI-loss contingency (ROADMAP)**
- PII CRF weights + labeled CSV are unrecoverable (Box-only) → stage will be rebuilt on Presidio + Indian-pattern recognizers behind the unchanged interface.
- Both at-risk HF models (`DPIC-Pipeline/vit_type_classifier`, `jeseniapar/categorizer` incl. label encoder) **mirrored into our DVC remote** — runtime no longer depends on accounts nobody controls. Model-provenance rule recorded.

## Test plan
- [x] 42 passed locally (pipeline structure/CLI/OCR-path, exporter roundtrip/idempotency/update on SQLite + Postgres, NUL-trackingId join regression) + ruff clean
[...]

PR #3, opened 2026-07-02
Week 2: CI Postgres service, mappings in DVC, compose oltp

ymohanty:
## Summary
Week-2 plumbing (draft — will retarget to main after #2 merges):
- CI: postgres:16 service on 5433 so the Postgres-path tests run instead of skipping
- janasunani-mappings master tables DVC-tracked (Phase 9 gate cleared)
- deploy/docker-compose.yml with the oltp service — adopted live on the box with zero data movement (external volume), backup path verified

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #40, opened 2026-07-22
docs: rebuild the plan around the five demo components for 14 August

ymohanty:
## What & why

Rebuilds the planning documents around the five demo components the ED set for **14 August 2026**, and grounds the intelligence layer on measurements taken against the real corpus rather than on assumption.

The branch is docs plus a small amount of document tooling. It changes no application code.

**Base is `deploy/cpu-box`.** Intentionally stacked on the deploy chain: `deploy/cpu-box` was caught up to `main` first, then merged in here, so this diff stays docs-only and merges through the deploy chain rather than racing `main`.

## What is in the diff

| Path | What |
|---|---|
| `docs/ROADMAP.md` | Engineering source of truth. Rewritten, ~2,500 lines changed |
| `docs/DELIVERY.md` | **New.** The management commitment for 14 August. Short, non-technical, with fallbacks and decision dates |
| `docs/ARCHITECTURE.md` | Trust tiers, the corrected PII boundary, status deduplicated |
| `docs/Full Technical Report DPIC.pdf` | **New.** The earlier DSI clinic's report. The only surviving record of the legacy benchmark numbers |
| `scripts/md_to_docx.py`, `Makefile` | **New.** Renders the delivery brief to Word through the `dpic` package's branded template. Normalises Markdown first, since that dialect has no list support |
| `janasunani/pipeline/README.md` | Fixes a contradiction with the roadmap on typed PII placeholders and duplicate similarity |
| `AGENTS.md`, `README.md`, `.gitignore` | Cross-references, generated-docx ignores |

The PII gold verifier that briefly lived here has moved to #54, where the labelling work it supports actually sits. History is intact; the files are simply not on this branch.

## Phase map

Part II, the demo (13–17), maps one-to-one onto the ED's five components:

| Phase | Component |
|---|---|
| 13 Pipeline completion: PII gold set, Presidio tuning, end-to-end run | (a) The DSI pipeline, rebuilt |
| 14 Spam & duplicate detection | (b) |
| 15 Structured analytics: metrics layer, spikes, closure, thematic clustering | (c) The intelligence layer |
| 16 A/B instrumentation + retrospective impact evidence | (d) |
| 17 Sarvam benchmark + provider registry + egress control | (e) |

Part III, post-demo (18–24), is a set of **gated investments, not an implementation commitment**: evaluation harness and operational safety, the model/pipeline platform, adjusted comparisons and natural-language querying, Odia-first models, semantic intelligence, the governed feedback loop, and the jurisdiction pack. §5 renumbers the old 13–19 map onto this; §2 records the crosswalk so old issue references still resolve.

## Two invariants that changed

**The "citizen text never leaves the box" rule is retired.** It was policy only, and it blocked the Sarvam benchmark the ED asked for. It is replaced by explicit trust tiers, `same-host` / `dpic-infra` / `authorized-external`, with exactly one module permitted to make an external call, a recorded authorization behind every such route, and a maintained lower-tier counterpart for each one so no capability becomes unrevocably dependent. Phase 17 owns the enforcement that was previously assumed.

**The lake was never PII-free, and the docs said it was.** `materialize.py` copies `SELECT *`, so `complaints.grievance` — the citizen's own account of the problem — sits in Parquet verbatim, having never met `pii_tagger`. §3.2 states what actually holds PII. The consequence is a scheduled build step: `janasunani-redact-grievance` runs before any index is built, and MinHash signatures and embeddings read only the redacted column. Those derived artifacts stay classified as personal data regardless, because distinctive phrasing re-identifies where a phone number no longer does.

## What the measurement work established

Phase 15 was re-grounded on aggregate queries against the 6.56M-row action history:

- Officers close cases on a graded set of standard phrases. **61% of the ~792,000 complaints closing on one of them use the wording that claims no action**, with the more specific wording available and used 311,000 times. The denominator is stated precisely, and the finding is framed as a description rather than a verdict: sometimes no action is correct, and a correct closure looks identical to a premature one in the record.
- `action_taken_remark` is 99.87% populated, and the top ten strings cover 45% of rows. Templating makes the action-type taxonomy tractable by hand.
- The bespoke free-text tail carries no hidden signal. The raw reopen-rate gap is confounded by case length and does not survive stratification.
- The ~34,700 officer-confirmed duplicates are a **baseline, not a deliverable**. The claim is the increment beyond them.

Each item in the intelligence layer now has to pass one test: could a competent analyst with SQL access build this in a day? What passes ships as a SQL view we hand over. What fails is the capability claim.

## Review focus

- §3.2 and whether the redaction path is now genuinely executable rather than merely documented.
- §5.6, the execution plan: what parallelises across agents, what is strictly serial, and the human bottlenecks. The hand-transcribed OCR sample is unassigned and is the largest unmanaged risk in the plan.
- Whether DELIVERY Table 1's Committed / Bounded / Framework-only split is defensible per component, and whether each fallback is one we would actually accept on the day.
- Phase 20's natural-language querying spec: semantic parsing into the semantic layer's schema, never text-to-SQL, and explicitly out of August.


ymohanty:
@codex review

Addressed the language-ID ordering finding: reframed Phase 13.1 into a pre-OCR **image-based** signal (for OCR-engine selection) vs a post-OCR **text-based** IndicLID refine + IndicXlit transliteration (or immediately on the typed-text path), and updated 13.2 / the Language-first invariant / ARCHITECTURE for consistency (commit above).

The two `scripts/setup.sh` P2 findings are valid but pre-existing code not touched by this docs-only PR — tracking them for a separate code PR rather than scope-creeping the roadmap change.

ymohanty:
@codex — separate from the line-level review, I'd like a **CTO-style architectural / product-design review** of this system and the Part III plan. Please read `docs/ARCHITECTURE.md` and `docs/ROADMAP.md` (esp. Part III: Phases 13.0/13/14/15) and evaluate at the *strategy* altitude, not line-by-line. Specifically:

1. **Overall architecture soundness** — the OLTP → Parquet-lake → pipeline split, the model-maturity spine (status-quo → Odia-first models → online learning) on the 13.0 modularity foundation, and the governance-intelligence layer. Is this the right shape? Where is it over- or under-engineered for a single-maintainer prototype heading toward a government/DPI-export product?

2. **Biggest architectural risks & what breaks first** — scaling (1.37M/6.56M rows, on-box models), the phased-hybrid local-LLM decision, the two-speed online-learning design, the human-in-the-loop correction loop, and the frozen serving contract as it accretes new endpoints.

3. **Strategic trade-offs & gaps** — evaluation rigor, observability, security/PII posture (the "citizen text never leaves the box" invariant), cost, and long-term maintainability. What's missing that a CTO would want before committing to this as a product?

4. **Sequencing** — is demo → improved-models → online-learning, with the intelligence layer as a parallel post-demo track, the right build order given the goal of attracting third-party governments?

Please be direct about what you'd change, cut, or resequence. Prioritize the few decisions that matter most.

ymohanty:
## CTO review — architecture and Part III strategy

**Verdict:** the core architecture is sound for the prototype, and the Part III direction is strategically credible. I would not, however, execute Phases 13–15 in the current order or at the current breadth. The largest risk is not whether DuckDB can scan 1.37M complaints or whether another model can be registered; it is whether the team can prove that multilingual outputs are safe and useful, operate the system reliably around real citizen data, and turn officer feedback into trustworthy labels. Part III should therefore be **evaluation-first, workflow-first, and platform-light**.

### What I would keep

- **OLTP → Parquet lake → resumable pipeline artifact DB is the right shape.** Postgres remains the transactional system of record; Parquet/DuckDB keeps analytics and training off production; the per-run SQLite DB is justified by resumability and the mutually incompatible ML environments. At this scale, the row corpus itself is not the problem.
- **Odia-first rather than translate-through-a-cloud API is the right product and privacy decision.** It respects the stated data boundary and creates a defensible capability for an Odisha-first system.
- **Batch and live inference must converge on one processing abstraction.** The roadmap correctly identifies drift between the batch pipeline and the warm processor as a structural risk.
- **The local LLM is correctly deferred.** Keep it as an offline, evidence-linked batch assistant for cluster narration or difficult cases—not a serving-path dependency.
- **Spike detection is a strong product wedge.** It is cheap, explainable, immediately useful to government, and less risky than learned routing or generative narration.

### The few changes that matter most

#### 1. Put evaluation before the registry/platform work

Phase 13.0 currently bundles four different investments: a generic stage registry, a model resolver, MLflow infrastructure, and promotion gates. Only two are essential now: a **single typed processing recipe shared by batch/live** and an **immutable model-release manifest with rollback**.

A freely configurable/reorderable stage DAG is over-engineered for six stages and creates a safety hazard: some orderings are policy invariants, not merely dependencies—PII-safe text must feed downstream summarization and external presentation, original-text offsets must remain authoritative, and batch/live semantics must match. Prefer a small number of named, versioned recipes with declared inputs/outputs and startup validation. Do not make arbitrary reordering a product feature.

Likewise, MLflow should be a control-plane tool, not a runtime dependency. Resolve an approved alias at deploy/startup, pin the resulting artifact/version in a release manifest, cache locally, expose it in health/telemetry, and retain a one-command rollback. “Hot-switching with zero code change” is useful; **unreviewed automatic production switching is not**.

Before either abstraction expands, build the per-language evaluation harness and gold-set governance. The surviving 80.56% PII “any overlap” score is a historical reference, not a release threshold. The minimum scorecard should include:

- PII precision/recall and false-negative rate by entity, language, document type, and source;
- OCR quality plus downstream task success, not just heuristic pass rate;
- categorization/routing top-k accuracy, calibration, abstention coverage, and override rate;
- summary faithfulness, omission, harmful disclosure, and officer acceptance;
- latency and cost by path/model/language.

Gold sets need immutable versions, annotation guidance, adjudication, and controlled access. Without this, the registry simply makes unsafe models easier to deploy.

#### 2. Reframe Phase 14 as governed feedback capture—not “online learning”

Officer edits are valuable, but they are not automatically ground truth. They may encode local practice, workload pressure, policy disagreement, inconsistent taxonomies, or malicious/accidental changes. A nearest-neighbour layer that immediately changes the next decision can amplify one bad correction and create cross-office or cross-jurisdiction leakage.

Ship this in three steps:

1. **Capture:** authenticated, append-only corrections with actor, reason code, before/after values, timestamps, model/release version, and audit history.
2. **Curate:** quality checks, deduplication, adjudication/approval, taxonomy-version compatibility, and explicit inclusion in a versioned lake-backed training set.
3. **Learn:** retrieval in shadow/suggestion mode first; periodic batch retraining only after a minimum clean-label volume and held-out evaluation; promotion via controlled rollout and rollback.

Call this a **human-feedback learning loop** until it demonstrably meets an online adaptation SLO. Keep routing decision support assistive: disposal time is confounded, and historical office assignment reflects the old policy. The roadmap’s OVB caveat is correct but should be elevated from a note to a release constraint.

The latest inline findings expose a real dependency issue here: the normalized semantic index must exist before the “fast” layer, and correction data must be materialized to the lake. Direct OLTP training should not be an allowed shortcut.

#### 3. Resequence the intelligence track around value and prerequisites

“Parallel” should not mean “independent.” I would order the post-demo work as:

1. Finish the demo baseline and operational safeguards.
2. Build gold sets, evaluation, observability, and a versioned release manifest.
3. Ship **B.2 spike/anomaly detection** using existing structured fields.
4. Implement language refinement/normalization and the smallest multilingual model improvements that clear the gates.
5. Build normalized embeddings and case retrieval, benchmarked on the actual CPU/GPU envelope.
6. Add governed correction capture and shadow-mode retrieval.
7. Retrain only when correction volume/quality justifies it.
8. Add topic induction and local-LLM narration last.

This makes the first post-demo feature an explainable governance outcome, not invisible ML plumbing. It also prevents embedding the same out-of-distribution text that caused the prior BERTopic failure.

The vector layer needs a capacity gate. Millions of high-dimensional vectors plus an HNSW graph are a materially different memory/IO workload from Parquet analytics on the downsized CPU box. Before choosing VSS/HNSW, benchmark artifact size, build time, filtered-query p95, memory, restart/rebuild behavior, and incremental updates. Brute force may be acceptable for offline analysis but should not be assumed acceptable for interactive correction retrieval. Model dimension, quantization, filtering/partitioning, and licensing should be chosen from the benchmark—not fixed by roadmap fashion.

### What will break first

In likely order:

1. **Evaluation and label quality.** The team will be unable to distinguish a better multilingual model from a more confident unsafe one.
2. **Operational safety around real data.** The current real-history path has no auth/redaction, the demo CORS default is permissive, backup automation is not reproducible from code, and the production database lives on a single-instance root-volume lifecycle.
3. **Human workflow adoption.** If correction entry is slow, unaudited, or has no visible benefit, officers will not produce the feedback signal the learning plan assumes.
4. **Batch/live and model-release drift.** Conflicting environments, remote model code, mutable aliases, and two execution paths can make a result irreproducible unless every output records the recipe/model/data versions.
5. **Embedding index footprint and latency.** The tabular corpus is modest; the dense semantic index is the first genuinely new scaling regime.

### Security and product gaps before calling this DPI-exportable

“Citizen text never leaves the box” is necessary but not sufficient. Embeddings, summaries, corrections, logs, and backups remain sensitive derived citizen data. Before a real deployment, I would require:

- identity/RBAC before any real history, correction, or intelligence endpoint; tenant/jurisdiction isolation; immutable audit logs;
- TLS, request/file-size limits, document validation/malware controls, rate limiting, retention/deletion policy, encryption at rest, backup encryption, and tested restore;
- network-egress enforcement or an explicit allowlist. Today the infrastructure permits general outbound access and some models are public runtime dependencies, so the no-egress rule is primarily policy rather than a technical boundary;
- mirrored, checksummed runtime artifacts; dependency/SBOM and model-license review; special scrutiny of **trust_remote_code**;
- stage-level metrics/traces: latency, failure, abstention, language distribution, redaction counts, model/recipe version, drift, and rollback events;
- API versioning (**/v1**), idempotency, structured errors, and asynchronous job/status semantics for long document/GPU work. Preserve the current contract as v1, but do not interpret “frozen” as “never evolve.”

For third-party-government adoption, the most important missing abstraction is not another model registry; it is a **jurisdiction pack**. Taxonomy, routing mappings, languages, model release, evaluation thresholds, retention rules, and RBAC policy should be configuration/data with a portability test against a second synthetic jurisdiction. Until that works, this is an Odisha deployment with reusable code—not yet a DPI product.

### Recommended scope cut

For the next milestone, I would explicitly defer:

- local-LLM narration;
- autonomous fast adaptation from individual corrections;
- learned routing optimized on disposal time/citizen benefit;
- a fully generic reorderable pipeline framework;
- HNSW unless measured interactive retrieval requires it.

I would fund instead: evaluation data, auth/audit, restore drills, observability, spike detection, Odia PII/routing quality, and a second-jurisdiction configuration exercise. Those are the capabilities that create trust and portability.

### Go/no-go

**Go** on the architectural direction and merge the plan once the outstanding sequencing/safety inconsistencies in the inline review are resolved. **No-go** on presenting Part III as an implementation commitment until four gates exist:

1. a reproducible, authenticated, observable demo baseline with a tested restore;
2. per-language/task release thresholds and versioned gold sets;
3. a normalized, lake-backed data path for embeddings and corrections;
4. governed correction curation with shadow evaluation before any adaptive behavior.

That narrower sequence keeps the good architecture, removes premature platform work, and maximizes the chance that the first post-demo investments produce government-visible value.

— **Codex**

ymohanty:
## CTO review addendum — after reviewing deploy/cpu-box

I reviewed **deploy/cpu-box at 7b7640c**, including all of **feat/demo-integration** through f760704 and the subsequent deployment-hardening commits. This **does materially improve my assessment of the current baseline**, although it does not change my Part III sequencing recommendation.

### What I would revise in the original review

The demo deployment plane is considerably more mature than the branch behind PR #40 made visible:

- API and frontend images are immutable commit-SHA releases, with base images pinned by digest.
- Models, mappings, and lake files are mounted read-only rather than baked into images.
- Caddy is the only public service; it provides TLS and fail-closed site-wide Basic Auth.
- The live processor fails closed on missing model artifacts/binaries and is health-gated through the proxy.
- The deployment uses a serialized, manual GitHub Actions workflow, environment-scoped secrets, GitHub OIDC for temporary AWS access, and temporary runner-IP SSH ingress.
- The box-side rollout records a last-known-good tag, checks disk headroom, enforces expand-compatible migrations, health-checks API and frontend, smoke-checks the proxy, automatically rolls back images and Caddy configuration, and re-verifies rollback health.
- The branch has unusually strong real-code-path regression coverage for the deployment control flow.

That is a competent **single-maintainer demo release process**. I would therefore withdraw the implication that ordinary deploy fragility or completely unauthenticated public history is likely to be the first failure. The proxy now protects the real-data surface, and the rollout/rollback design is thoughtful.

My revised operational verdict is:

- **Go for the controlled demo deployment**, once the first real amd64 build and on-box browser E2E pass.
- **Still no-go for a government product deployment** without the remaining controls below.
- The first strategic failure remains **evaluation/label quality**, not infrastructure scale.

### What remains unchanged—or is now more precise

1. **Recovery is still the largest infrastructure risk.** The production database, lake, model cache, images, and backups share one host/root-volume failure domain. The nightly backup script and cron are still box-only and not reproducible from code, and the runbook does not demonstrate a restore drill. Codify the timer, retention, encryption, and restore test before Part III infrastructure.

2. **Code deployment is reproducible; inference releases are not yet fully reproducible.** An image SHA does not pin the independently bind-mounted DVC models, routing mappings, Parquet snapshot, schema revision, or the public Hugging Face BART revision/cache. Phase 13.0 should first produce an immutable release manifest joining all of those hashes/versions. That is more valuable than a generic reorderable stage framework.

3. **Basic Auth is right for this demo, not for Phase 14/15.** It protects the perimeter but supplies no officer identity, roles, jurisdiction/office scope, or per-action audit context. Governed corrections and intelligence views need SSO/OIDC-style identity, RBAC, append-only audit records, and tenant/jurisdiction isolation before they are exposed.

4. **The “citizen text never leaves the box” rule is still policy-heavy rather than technically enforced.** The host permits broad outbound traffic, BART is downloaded from a public hub at startup, and the box now stores a read-packages GHCR PAT. The latter also makes the current “boxes hold no GitHub credential” documentation inaccurate. Mirror and checksum every runtime model, define/limit egress, and either document and rotate the scoped PAT or move image pulls to an instance-role-native registry.

5. **Readiness is narrower than product readiness.** The rollout proves that the processor is warm and the frontend responds, but it does not prove that real history is queryable, the persistent result store is selected, routing mappings are loaded rather than silently falling back, or a submit→persist→fetch transaction succeeds. The branch itself lists the on-box browser E2E as unrun. Add a non-PII synthetic canary/readiness check that covers those dependencies and reports the release manifest.

6. **The CPU deployment artifact is unnecessarily heavy.** The documented 8–12 GB API image includes CUDA Torch even though DeepSeek is deliberately excluded and the always-on demo box is CPU-only. A CPU-only Torch build would materially reduce CI time, pull time, disk pressure, rollback time, and the chance that deployment affects the production database through shared-disk exhaustion.

7. **Two known rollout gaps should be closed before calling Phase 12 done.** The workflow can time out and kill the SSH session before the script completes its own health wait/rollback, and a rollback run can ship current compose/proxy artifacts beside an old image SHA. Detach/poll the box-side deployment and ship artifacts from the same commit as the requested image.

### Updated priority order

Given this branch, I would now sequence the work as:

1. Complete the first real CI build/on-box E2E and close the two rollout gaps.
2. Codify backup plus restore, add full dependency readiness, mirror BART, and shrink the CPU image.
3. Merge a consistent documentation baseline.
4. Build per-language gold sets/evaluation and observability.
5. Ship structured spike detection, then the minimum multilingual improvements that clear the gates.
6. Continue with normalized embeddings and governed human feedback as described in the original review.

The deployment work means the project has earned the right to stop investing in demo plumbing soon. The next marginal engineering week should go to **proof of model safety/usefulness and recoverability**, not more deployment abstraction.

### Important PR topology issue

PR #40’s description says its base is **deploy/cpu-box**, but GitHub currently reports the base as **main**, and the local **docs/roadmap-multilingual-intelligence** branch is not a descendant of **deploy/cpu-box**. That explains why the documents I originally reviewed still described only the OLTP Compose service and unauthenticated real-history caveats.

Before merging PR #40, either retarget/rebase it onto **deploy/cpu-box** or merge the deployment branch first and then rebase the docs PR. Otherwise the source-of-truth roadmap/architecture can regress or conflict with the actual Phase 12 state.

**Bottom line:** deploy/cpu-box raises my confidence in the team’s operational discipline and makes the prototype architecture more credible. It does not weaken the central CTO recommendation: Part III should remain evaluation-first, recovery-aware, and cautious about automatic human-feedback adaptation.

— **Codex**

ymohanty:
Pushed a full restructure of `docs/ROADMAP.md` (and matching `ARCHITECTURE.md` updates) that folds in both Codex reviews.

**Readability / dedup (per maintainer request):** the roadmap was a multi-session ramble with status in four places. It's now: one canonical phase/status table (§2, the only place status lives), an architecture-in-brief that points to `ARCHITECTURE.md` instead of restating it, tight per-phase descriptions with no status markers, the eval-first Part III, and a compacted decisions log. Dropped: the week-by-week narrative, the duplicated snapshot status, and the resolved/struck-through open-items + Reviewer-input blocks.

**CTO review folded in** — Part III is now sequenced evaluation-first, workflow-first, platform-light (Phases 13–19 in priority order): eval + gold sets + operational safety first; minimal modularity (fixed stage order + model registry + MLflow control-plane with a release manifest, **no** reorderable DAG); spike detection before the model work; Odia models gated on eval; embeddings normalization- and capacity-gated; governed feedback (shadow-first, OVB as a release constraint); jurisdiction pack; and the four go/no-go gates.

**Line-level findings** — the earlier A–H inline comments are resolved: #1/#C dissolve under the fixed-order recipe; #2 (transliteration offsets) is a stated privacy constraint; #3/#A MLflow aliases; #4/#E corrections are lake-backed; #5/#H VSS capacity-gated; #6/#G task-keyed eval + routing gold; #7 embedding source/kind; #B/#D dependency ordering baked into the sequence.

**deploy/cpu-box addendum folded in** — Phase 12 status now reflects the built+reviewed deploy pipeline; Phase 13 reconciles what it already down-pays (basic_auth, health-gated auto-rollback, scoped CI IAM) vs what remains (RBAC, egress, restore, audit); Phase 14's release manifest now joins all bind-mounted artifact hashes; added the CPU-only image, the two rollout gaps, backup/restore codification, the GHCR-PAT egress note, and the readiness canary.

**Merge order:** this branch descends from `main`, which does not yet contain the `deploy/cpu-box` Phase 12 work (its PR chain, #29 → `feat/demo-integration`, lands on `main` separately). To avoid a `ROADMAP.md` collision, **the deploy work should land on `main` first**, after which I'll rebase this PR and flip Phase 12 from "unmerged" to "merged." Please don't merge this ahead of the deploy chain.

ymohanty:
## Final CTO refinements after the roadmap rewrite

The rewritten roadmap now conforms to roughly **90–95%** of my expectations. I approve the strategic direction: it is evaluation-first, workflow-first, and platform-light; removes the generic reorderable stage DAG; treats MLflow as a deploy/startup control-plane; prioritizes a full release manifest; ships explainable spike detection before semantic indexing; makes feedback curated and shadow-first; and introduces the jurisdiction pack as the DPI portability mechanism.

I would make five final refinements:

1. **Allow Phase 15 to run alongside Phase 14.** Spike/anomaly detection uses structured lake fields and does not depend on the model registry or shared inference recipe. Once Phase 13’s safety/evaluation foundation exists, Phase 15 can deliver the first government-visible value while Phase 14 prepares the model platform needed by Phase 16. Avoid serializing them unnecessarily.

2. **Define the jurisdiction boundary earlier.** During Phase 14, establish the minimal configuration contract for taxonomy, routing mappings, languages, retention, RBAC, and evaluation thresholds. Keep the second synthetic-jurisdiction implementation/validation in Phase 19. Otherwise Phases 15–18 may embed Odisha-specific assumptions that are expensive to extract later.

3. **Add an explicit API evolution policy.** Preserve the current response contract as v1, but specify additive compatibility, versioned routes, idempotency, structured errors, upload/request limits, and asynchronous job/status semantics for long OCR or GPU operations. “New schemas; frozen contract untouched” will not be sufficient as corrections and intelligence endpoints accumulate.

4. **Describe Phase 14’s shared processor as a fixed typed dataflow with explicit branches, not literally one linear stage sequence.** Typed grievances skip image classification, OCR, and page-type work; document submissions include page-level processing and gating. The important invariant is one validated batch/live recipe with explicit text/document branches and mandatory safety edges—especially original-text offsets and redaction before summarization/presentation.

5. **Resolve the remaining Architecture inconsistencies.** ARCHITECTURE currently says the project has “two parts,” although Part III is now planned, and says the CPU box “will grow api/frontend/mlflow/proxy services.” It should say that api/frontend/proxy are built on deploy/cpu-box but not yet merged/applied, while MLflow remains planned for Phase 14.

Two smaller wording adjustments:

- Replace “before anything touches real data” with “before any additional real-data exposure or Part III endpoint”; the system already stores and processes real data.
- Treat BGE-M3 and the larger embedding models as **benchmark candidates**, not a predetermined floor/ceiling. Selection should follow multilingual task quality, licensing, artifact size, build cost, and measured CPU/GPU performance.

These are refinements, not strategic objections. The new roadmap is disciplined, candid about its gates, and appropriately cautious for a single-maintainer prototype handling citizen PII.

— **Codex**

ymohanty:
Folded in all five final refinements + both wording tweaks (commit 88c9714):

1. **Phase 15 alongside Phase 14** — noted in the execution order; spike detection uses only structured lake fields, so it isn't serialized behind the model platform.
2. **Jurisdiction contract early** — Phase 14 now establishes the minimal config seam (taxonomy/mappings/languages/retention/RBAC/thresholds); the second-jurisdiction *validation* stays in Phase 19.
3. **API-evolution policy** — added to Phase 14: pin the contract as v1, additive compatibility, versioned routes, idempotency, structured errors, size limits, async job/status for long OCR/GPU work.
4. **Typed dataflow, not a linear chain** — Phase 14's shared recipe is reframed as a fixed typed dataflow with explicit text/document branches and mandatory safety edges (redaction before summarization/presentation; original-text offsets authoritative).
5. **ARCHITECTURE inconsistencies** — now "three parts" (Part III added); the CPU-box line reads api/frontend/proxy built-on-`deploy/cpu-box`-but-unmerged, mlflow planned for Phase 14.

Plus: "before any *additional* real-data exposure" (the system already processes real data), and embedders framed as **benchmark candidates** rather than a fixed floor/ceiling.

Merge order unchanged: `deploy/cpu-box` lands first, then I rebase this and flip Phase 12 to merged (at which point the CPU-box line takes the deploy branch's present-tense wording). Thanks for the review.

ymohanty:
## CTO follow-up review — latest ROADMAP / ARCHITECTURE revision

**Verdict:** yes, this now makes sense as a product and technical strategy. The latest revision materially improves the plan and conforms to the direction I expected from the earlier review. I would approve the architectural direction, with a few execution refinements before Phase 15 is treated as implementation-ready.

### What is now right

- **The structured/unstructured split is the correct boundary.** Structured analytics over category, district, dates and action history can produce government-visible value without waiting for Odia normalization, embeddings or a model platform. Text retrieval and theme discovery correctly remain later, because they introduce language, privacy and capacity risk.
- **The dependency graph is much healthier.** Evaluation and operational hardening remain first; the model platform stays deliberately small; Phase 15 can run alongside Phases 14/16; and the grievance-processing path does not depend on an interactive LLM.
- **The local-model decision is appropriately reversible.** A capacity-gated query model with a guided/templated fallback is the right posture for the CPU box. The plan does not make an external API or an always-on GPU a hidden production dependency.
- **Adjusted comparison is better than a blanket “never compare offices” rule.** Government users will ask comparative questions. Encoding approved measures in a governed metrics layer is a stronger answer than pretending the question will not arise.
- **ROADMAP and ARCHITECTURE are now aligned at the strategic level.** The fixed typed processing flow, release manifest, early jurisdiction boundary, API evolution policy, governed feedback loop and evaluation-first sequencing form a coherent whole.

### Refinements I would make

1. **Make Phase 15 an explicit sequence, not one large promise.** The phrase “highest-ROI, lowest-risk” is true of the structured-data opportunity, but not equally true of every item now inside it. I would sequence it:
   1. tested metric definitions, data-quality/freshness reporting and fixed dashboards;
   2. deterministic spike detection and contribution analysis;
   3. adjusted comparisons, only after statistical and policy review;
   4. natural-language querying, only after the first three are trusted.

   This preserves early visible progress and prevents the query agent from becoming the critical path.

2. **Add explicit query and disclosure controls.** “Constrained SQL” and RBAC are necessary but insufficient around citizen records. Before implementation, the roadmap should require: an allowlisted query plan/AST rather than arbitrary SQL; a read-only isolated execution role/process; time, memory and result-size limits; minimum-group-size suppression for small cells; restricted drill-down fields; prompt/query/result audit logs; and visible data-source/freshness provenance on every answer. Otherwise a valid query can still disclose too much or exhaust the shared CPU box.

3. **Qualify “language-agnostic.”** Phase 15 does not require reading or normalizing grievance text, which is the important architectural point. But its category and routing fields may still inherit missingness, historic policy choices and language-related classification error. I would say “does not depend on text-language processing” and require profiling/reconciliation of the structured fields before adjusted comparisons are exposed.

4. **Turn the strongest technology claims into hypotheses.** Statements such as “SOTA route,” a local LLM reading this material “far better than any embedder,” and a capability that “no off-the-shelf BI tool offers” are more confident than the evidence in the roadmap supports. The document’s evaluation-first philosophy is stronger: benchmark these options and promote them only if they win on multilingual quality, privacy, cost and operational reliability.

5. **Refresh the PR description before merge.** It still describes the old Phase 13.0/13/14 structure, Phase 15 as the embedding/theme track, and says the base is `main`; the actual PR now targets `deploy/cpu-box` and the roadmap has Phases 13–19 with a different sequence. The description should match the decision being approved.

One wording inconsistency also remains: Phase 13 says operational safety is required “before any real-data endpoint,” while the execution-order text correctly says before any **additional** real-data exposure or Part III endpoint. The latter is accurate because the system already stores and processes real data.

**Bottom line:** the plan is now strategically coherent and appropriately ambitious. Keep the two-track intelligence architecture, but make deterministic structured analytics the first shipped slice, treat adjusted/agentic analytics as gated increments, and specify the query/privacy controls before implementation. With those refinements, I would be comfortable using this roadmap as the basis for funding and sequencing the next phase.

— **Codex**

ymohanty:
## CTO follow-up review — latest ROADMAP / ARCHITECTURE redraft

> **Review scope note:** PR #40 currently points at commit `38f1e70`. The newest `ROADMAP.md` and `ARCHITECTURE.md` redraft I reviewed is still uncommitted in the local working tree, so this review anticipates the next push. Please re-check the final diff after those files are committed.

### Verdict

The new direction makes strategic and architectural sense. I would approve the **direction**, but I would not yet approve the document as a credible **August 14 delivery plan**.

The strongest improvements are:

- The demo is now anchored to five explicit outcomes, with an honest distinction between working features and framework/evidence deliverables.
- Spam, repeat submissions, and public campaigns are correctly treated as different phenomena.
- Model accuracy and real operational benefit are separated — an excellent distinction.
- Sarvam is positioned as something to benchmark, with reversibility and fallback, rather than adopted on faith.
- Structured analytics is sensibly decoupled from the harder Odia-language work.
- The architecture diagram and phase crosswalk make the substantial re-scope understandable.

### Material refinements required

#### 1. Add an actual August 14 delivery plan

There are no dates, owners, resource assumptions, weekly milestones, or critical-path table in either document. Yet the remaining demo scope includes unfinished deployment plus Phases 13–17. Important dependencies include Phase 13 → Sarvam evaluation and Phase 14 → Phase 15.

For August, I would narrow the promised results to:

- a working end-to-end pipeline and privacy scorecard;
- advisory spam detection and deterministic duplicate/campaign identification;
- three to five fixed management metrics and one spike-detection view;
- an experiment design, power calculation, and minimum exposure-log schema — not necessarily a complete experimentation platform;
- a hosted Sarvam adapter and benchmark scorecard. Defer the MLflow service and operational self-hosting of the 30B model unless capacity already exists.

Add a table with: **deliverable, owner, target date, dependency, present status, acceptance test, and fallback**.

#### 2. Redefine the security boundary

“Leaves the box” is not technically coherent when documents already move through S3 and the proposed self-hosted Sarvam model runs on another GPU machine. The architecture also classifies the GPU route as `on-box`.

Define a trust boundary instead:

- same host;
- DPIC-controlled infrastructure, including S3 and GPU machines;
- authorized third party.

Each route should declare the data class, destination, approval reference, retention terms, encryption, audit policy, and fallback — not only `on-box` versus `authorized-external`.

#### 3. Correct the lake/PII contradiction

Phase 13 says no unredacted PII reaches the lake, but the proposed duplicate index derives keys from mobile/email fields obtained from the lake. The lake is therefore not uniformly PII-free.

State precisely which lake fields contain personal information, which roles may access them, and — if this is the intended guarantee — that no **new unredacted grievance text** reaches downstream outputs.

#### 4. Do not state categorically that “no IRB applies”

An operational evaluation may be exempt, but that should be a recorded determination by the appropriate government/legal/ethics authority, not an architectural assumption. ICMR guidance says programme evaluation can sometimes be exempt, while evaluation intended to produce generalizable knowledge requires proper ethical review: [ICMR National Ethical Guidelines](https://ethics.ncdirindia.org/asset/pdf/ICMR_National_Ethical_Guidelines.pdf) and [Department of Health Research guidance](https://www.dhr.gov.in/offerings/schemes-and-services/details/national-ethics-committee-registry-for-biomedical-and-health-research-necrbhr-UTNyUTMtQWa).

Likewise, replace “no arm may leave a citizen worse off” with enforceable protections: retain the existing service path, monitor harm indicators, define escalation, and establish predetermined pause conditions.

#### 5. Preserve campaign volume as well as unique-issue counts

The roadmap says campaigns are valuable governance signals, but then requires spike detection to use deduplicated counts. A campaign of 500 citizens is not necessarily a false spike.

Report three measures:

- total filings;
- unique grievance clusters;
- unique citizens or signatories, using protected identifiers.

Label whether a spike is campaign-driven rather than suppressing it.

#### 6. Resolve the internal inconsistencies

- Phase 13 is marked not started, while its privacy evaluation set is explicitly in progress.
- ROADMAP Phase 17 says to add MLflow to Compose; ARCHITECTURE says MLflow remains Phase 19.
- The Sarvam security cross-reference says §5.4; it should be §5.5.
- ARCHITECTURE says status belongs in ROADMAP but creates another status table. Change that table to “architectural impact” so it cannot drift.
- ARCHITECTURE calls cloud infrastructure “built,” while also saying the live stack has not yet been brought up.

One additional implementation detail: replacing every phone number or email with the same typed placeholder can artificially increase MinHash similarity. Exclude or down-weight generic PII placeholders for text similarity, while using separately salted identity keys for same-citizen resubmissions.

### Does the plan document read cleanly?

It reads clearly and thoughtfully for a senior technical reader, but it is **not yet a clean management plan**. At roughly 1,000 lines, `ROADMAP.md` is simultaneously a roadmap, technical design, research protocol, risk register, and decision log. It also retains specialist language such as “gold set,” OLTP, MinHash/LSH, SUTVA, semantic layer, AST, and HNSW.

I would keep it as the engineering source of truth, but put a two-page executive delivery plan at the front — or maintain one separately — containing:

- **Done / ongoing / scheduled** written in words, not unexplained symbols;
- the five promised outcomes;
- August dates and owners;
- visible progress for each outcome;
- major dependencies and fallback scope;
- decisions required from management;
- risks expressed in plain language.

So: strategically strong, technically serious, and much better aligned with the requested demo. The remaining issue is less “does the architecture make sense?” and more “does this document prove that this particular team can deliver the promised August outcome?” It does not quite do that yet.

— **Codex**

ymohanty:
## CTO follow-up review — ROADMAP, DELIVERY, ARCHITECTURE and supporting changes

**Reviewed at PR head `a38c0e9`.** I also checked the newly committed DSI technical report against the benchmark figures and reviewed the added PII-verification and Markdown-to-Word tooling.

### Verdict

This is materially better than the previous version. The architectural direction is now coherent, the trust boundary is honest, the lake/structured-PII distinction is substantially clearer, campaign volume is no longer incorrectly erased, and `DELIVERY.md` is exactly the right idea: a short management commitment sitting above the long engineering source of truth.

My verdict is **conditional approval**. I would approve the architecture, but I would not yet sign off the complete 14 August promise. Four delivery-critical contradictions remain, followed by several smaller consistency and tooling fixes.

### What this revision fixed well

- The single-engineer constraint and 14 August date are now explicit.
- The management commitment is separated from full phase scope, with fallbacks and decision dates.
- The security model now uses `same-host` / `dpic-infra` / `authorized-external`, which is much more defensible than “on the box.”
- The structured contact fields in the lake are acknowledged as personal data.
- Campaigns now retain filings, distinct-problem and signatory counts rather than being collapsed into a false single truth.
- Phase 13 status, Phase 17 MLflow ownership, architecture status duplication, and the earlier cross-reference problems are corrected.
- The intelligence section distinguishes a cheap insight from a capability that genuinely needs new processing. That is commercially and politically important.
- Natural-language querying is properly out of August and has a credible post-demo architecture.

## Delivery-critical findings

### 1. The “before and after” benchmark promises evaluation data that is not scheduled

`DELIVERY.md` promises that page type, category assignment and summarisation will all be re-measured by 14 August, and says every stage will be scored against the DSI clinic’s numbers. But the only current labelled set described for August is the 85-page privacy set. The roadmap explicitly leaves per-task/per-language evaluation for every other task to Phase 18.

The DSI PDF confirms why the existing figures cannot simply serve as the “before” half of a comparable before/after table:

- OCR’s 77.89% is a heuristic plausibility pass rate on 96,469 English DeepSeek pages, **not OCR accuracy**; there was no transcription ground truth.
- Page-type accuracy came from the DSI team’s own 1,500-page labelled sample.
- Categorisation came from their particular train/test split.
- Summarisation used a 0–3 qualitative scale assigned by one reviewer over 500 pages.
- PII is the only August task with a newly described human-labelled set.

So rename Table 2 from **“Before and after”** to **“Historical reference and current measurement”**, and add an evidence manifest for each row: sample, sample size, languages, label source, reviewer, metric and whether comparison is genuinely like-for-like.

For August, I suggest committing only to measurements for which labels actually exist:

- PII on the corrected 85-page set;
- OCR/Sarvam Vision on a deliberately transcribed printed/handwritten sample;
- duplicate recall against the officer-confirmed set;
- categorisation against an explicitly identified held-out label set, if one is available;
- summarisation only if a blinded human review can be scheduled. Otherwise present the DSI number as historical context, not a re-measured row.

Without that change, the benchmark table is a promise to produce evidence for which the plan contains no evidence-production step.

### 2. Phases 14 and 15 secretly depend on Phase 17, although Phase 17 is scheduled later

The stated one-engineer order is `13 → 14 → 15 → 17 → 16`. But Phase 14 requires romanised Odia to be transliterated through the Phase 17 provider before hashing, and Phase 15 S4 says “transliterate first, then embed.” Both therefore depend on work scheduled after them.

Choose one honest August contract:

1. pull a minimal transliteration adapter ahead of Phase 14; or
2. ship script-specific lexical matching in August and explicitly report cross-script recall as unsupported/stretch.

The management document currently overclaims further: it says duplicate-adjusted workload matches complaints “written in different words across three scripts,” while ROADMAP correctly says semantic duplicates — same issue, different words — are **not handled by Phase 14**. MinHash is lexical near-duplicate matching, not same-meaning matching. Change the management wording to “same or near-identical wording, with cross-script matching only if transliteration lands.”

The same scope issue affects thematic clustering. Reusing the existing MuRIL encoder avoids a new model, but running it over 1.37 million historical records is still a new corpus-scale inference and clustering job. It is not a free “forward pass we already make.” For August, make full-corpus S4 a stretch goal or bound it to one priority category/time window with a precomputed demo artifact.

### 3. The PII table is still wrong about historical grievance text

The roadmap says the Parquet lake’s text fields contain only redacted text. The implementation materialises `SELECT *` for the entire `complaints` table, and `complaints.grievance` is the citizen’s historical grievance text. Phase 15 then explicitly proposes reading that field over the full corpus.

Therefore the lake contains more than structured contact PII: it also contains raw historical grievance text unless a separate transformation not described here redacts it. The narrower guarantee “no un-redacted grievance text reaches downstream output” is also ambiguous when that text is already a downstream Parquet input.

The documents should say one of the following, whichever is true:

- historical `complaints.grievance` is raw personal data in the lake, access-controlled and redacted before MinHash/embedding/analytics; or
- materialisation is changed to produce a separately redacted analytics field and the raw field is excluded from analytics-facing Parquet.

Also classify MinHash signatures and embeddings derived from raw grievance text as governed personal-data artifacts. A signature/vector is not automatically anonymous merely because it is not readable prose.

Until this is resolved, Phase 14 and Phase 15 should not claim their derived indexes are downstream-safe by construction.

### 4. The one-engineer schedule remains overcommitted

`DELIVERY.md` is much cleaner, but Week 2 still contains: privacy scoring, a new corpus-level pipeline stage, spam logic, duplicate validation, a 1.37M-record backfill, semantic-layer definitions, management API/UI, thematic embeddings/clustering and a spike view. Week 3 then carries Sarvam across multiple tasks, a full benchmark table, experiment design and a power calculation.

That is not a credible full commitment for one engineer in the remaining working days, especially because deployment and the first end-to-end run are not yet complete.

I recommend a visible RAG/status column and per-component fallback in Table 1:

- **(a) Must ship:** end-to-end live run plus PII scorecard.
- **(b) Must ship, bounded:** lexical duplicate/campaign detection on a defined corpus slice; spam may fall back to deterministic warnings only.
- **(c) Must ship, bounded:** closure SQL view plus one precomputed, reconciled three-count/spike example; full thematic clustering is stretch.
- **(d) Framework only:** analysis plan and power calculation; no implication that trial instrumentation is production-ready.
- **(e) Must ship, bounded:** Sarvam Vision and PII/OCR comparison on the labelled sample; 30B categorisation/summarisation and provider registry are stretch unless their evaluation data is already identified.

Map each fallback to its component rather than placing four fallbacks in one final sentence. That makes the management promise legible before something slips, not afterwards.

## Additional corrections before merge

### The closure finding is still stated too causally

“What separates [correct and premature closure] is whether the citizen returns” is too strong. Return is a useful risk signal, but it does not identify causal correctness: citizens may return after an appropriate decision, and may fail to return after a bad one. The roadmap itself says 300–500 adjudicated closures are needed to calibrate the claim.

For August, say: **return/reopen identifies a subset of no-action closures warranting review; it does not determine that the closure was wrong.** Also define the denominator behind “roughly 60%” precisely — the roadmap describes this as a share of explicit disposal-template selections, while the management text risks sounding like 60% of all closed cases received no action.

### Reconcile the current-state statements

- “The document pipeline is complete” conflicts with “not yet run start to finish.” Say “all six stages are implemented; end-to-end integration is unverified.”
- `DELIVERY.md` says the interface is built and connected, while ROADMAP says live-API wiring is still in progress.
- `DELIVERY.md` says routing uses fixed rules, while ROADMAP states the demo ships with `method:"fallback"`.
- The pipeline README says typed PII tokens preserve duplicate similarity, while ROADMAP correctly explains that common `[PHONE]` / `[NAME]` placeholders can inflate unrelated similarity and must be removed/down-weighted.
- The schedule says dedup must precede spikes to stop a campaign registering as a surge. A campaign **is** a real surge under the revised three-count design; dedup is required to explain whether it is one issue or many, not to suppress the spike.

### Fix the PII verifier’s completeness checks

The focused tests are green (**17 passed**) and targeted Ruff checks pass, but the verifier has three correctness gaps relevant to the August scorecard:

1. draft/gold record-set differences are reported but do not fail the command, so omitted pages can pass verification;
2. it does not verify that text is identical for a shared record ID, so edited text can invalidate offsets without detection;
3. `records_touched` is incremented only from residual additions/removals, so a record changed solely by re-bounding or relabelling can be reported as untouched.

Add tests for all three. I would also keep `--show-samples` out of any automated or shared workflow because it deliberately prints citizen PII.

### Refresh the PR description

The PR body is now materially stale. It still says the change is docs-only, describes the old Phase 13–19 map, and asserts the retired “citizen text never leaves the box” invariant. The PR now includes Python scripts, tests, a PDF, a delivery plan, trust-tier changes, and Phases 13–24. This must be updated before merge so the review and audit record describe what is actually being merged.

### Correct Sarvam rate limits

The price and Vision limits match the current official documentation, but the LLM tier table does not. Current Sarvam documentation lists chat-model limits of **40 / 60 / 120 requests per minute** for Starter / Pro / Business, not 60 / 200 / 1,000: [Sarvam rate limits](https://docs.sarvam.ai/api/getting-started/ratelimits). The 30B 64K context, default reasoning caveat, ₹0.50/page Vision pricing and 10-request/minute Vision limit are otherwise consistent with the current docs: [Sarvam-30B](https://docs.sarvam.ai/api/getting-started/models/sarvam-30b), [Sarvam Vision](https://docs.sarvam.ai/api/getting-started/models/sarvam-vision), [pricing](https://docs.sarvam.ai/api/getting-started/pricing).

### Record settled governance decisions rather than writing “do not re-open”

If the programme-evaluation/ethics determination and Sarvam authorization are settled, record the decision owner, date, scope and reference. “Settled; do not re-open it” is brittle wording in an engineering source of truth. The document should preserve traceability without inviting engineers to reinterpret the decision, but it should also allow a future reader to understand who made it and what change in scope would invalidate it.

## Readability of `DELIVERY.md`

This is now **substantially cleaner and appropriate for a non-technical manager**. The sequencing, plain-language explanation, explicit decisions and rehearsal freeze are all strong. It is the first version I would be comfortable circulating after the factual/scope corrections above.

To finish it:

- add **status / confidence / fallback** to Table 1;
- use “historical reference/current measurement,” not “before/after”;
- remove the cross-script/different-words overclaim;
- qualify the closure/return result as a review signal;
- replace “complete” with “implemented but not yet integrated” where appropriate.

The long ROADMAP is now an effective engineering knowledge base, but at 1,685 lines it is not itself a management plan. That is fine now that `DELIVERY.md` exists. Keep the two audiences separate and do not copy the roadmap’s specialist language back into the delivery brief.

**Bottom line:** the system design is now strong and the management document is close. The remaining work is to make the August promise match the evaluation data, dependency order, actual data boundary and one-engineer capacity. Once those four are reconciled, I would sign off the plan.

— **Codex**

ymohanty:
## Concise CTO re-review — latest revision

**Reviewed at PR head `e7795b5`. Verdict: approve with a few targeted corrections.**

This revision resolves most of my earlier concerns. The benchmark now distinguishes historical figures from measurements that can actually be produced; cross-script matching and thematic analysis are honestly bounded; the closure finding is properly qualified; the Sarvam limits are corrected; and `DELIVERY.md` now gives management a clear commitment, fallback and decision table.

I also revise my earlier capacity assessment. This is not a conventional solo-engineer workflow: it is **one accountable engineer, substantially augmented by autonomous frontier-model agents such as Codex and Claude Opus** across implementation, testing, analysis, documentation and independent review. With that parallelism, the August scope is credible—provided integration, real-data validation and final acceptance remain protected human-owned work.

Remaining issues:

1. **Make the historical-grievance privacy path executable, not just documented.** Section 3.2 correctly acknowledges that `complaints.grievance` is raw citizen prose and bypasses `pii_tagger`. But Phase 14 and the execution diagram still imply `pii_tagger → spam_duplicate → backfill`. For the historical corpus, explicitly schedule either a separate redaction pass before MinHash/embedding, or the access/retention/audit controls required if those derived indexes remain governed personal-data artifacts.

2. **Correct the Sarvam dependency wording and assign the transcription owner.** `DELIVERY.md` says the 85-page privacy correction gates the Sarvam benchmark, but the committed Sarvam Vision comparison is actually gated by the separate hand-transcribed OCR sample. That sample remains unassigned and is the principal human critical-path risk. Assign it by 31 July or invoke the already-written comparative-only fallback.

3. **State human accountability in the agent workflow.** Replace “one engineer, plus subagents” with the clearer operating model above, and state that the engineer gives final acceptance to PII-, security- and deployment-sensitive changes. Agent review should strengthen—not replace—the accountable approval.

4. **Refresh the PR record before merge.** The PR description still says docs-only, uses the old phase map and asserts the retired no-egress rule. It no longer describes the change being reviewed.

Subject to those corrections, the roadmap, architecture and delivery plan now form a coherent and defensible plan for 14 August.

— **Codex**

PR #58, opened 2026-07-28
Demo stack integration: live inference API, frontend wiring, Makefile bring-up

ymohanty:
## Summary

Integration branch for the demo stack. Collects the work already reviewed and merged in #26 (live inference API) and #27 (Makefile bring-up + frontend wiring), so it can reach `main` as one hop.

Opened as a draft to unblock the merge path: #29 (`deploy/cpu-box`, green and Codex-clean) is based on this branch, not on `main`, so 52 commits currently have no route to `main` without it.

15 commits, 18 files, +1,219/-213.

## What Changed

- **Live inference bring-up** (#26). `demo` uv extra plus a preflight that fails closed when tokenizer vocab or model weights are missing, a scoped `dvc pull`, and `janasunani/inference/README.md`.
- **Frontend on the live API** (#27). Submit and history routes read the real serving endpoints instead of mocked responses; `frontend/.env.local.example` and the frontend README follow.
- **Makefile demo targets** (#27). `make up` and friends wrap the `docs/DEMO.md` runbook, with a frontend off-switch and a self-provisioning DB.
- **Docs.** `docs/DEMO.md` added; README §7, the docs index, `docs/DEPLOY.md` and `docs/ROADMAP.md` point at the demo extra. Roadmap Phase 9 reframed as a purely learned scorer (tabled past the demo, #33).
- **Tests.** `tests/test_inference_live_builder.py` extended for the preflight and artifact requirements.

## Validation

- [x] `uv run ruff check .` — all checks passed
- [x] `uv run --extra serving --extra pipeline-core pytest` — 227 passed, 7 skipped
- [x] `cd frontend && npm run lint && npm run build` — clean, 3 routes prerendered
- [x] Notebooks have been stripped with `uv run nbstripout --install` or equivalent.
- [x] No proprietary data files are committed directly to Git.
- [x] DVC pointers / `dvc.lock` are updated when approved data or pipeline outputs changed.

Re-run against this exact merge state, with the heavy extras CI does not install. CI is separately green on all four checks (`no-raw-data-in-git`, `ruff`, `test-and-validate-pipeline`, coverage). Every commit was also reviewed on #26 or #27 at merge time.

## Review Expectations

- Keep this PR small enough to be reviewed by a human in a short sitting.
- Split unrelated changes into separate PRs before requesting review.
- Two-person review is required.
- Yashaswi is a mandatory reviewer.
- Add one additional reviewer who can assess the code, analysis, or domain impact.

[...]

Issue #49, opened 2026-07-27
Phase 13 · Pipeline completion (PII gold set, Presidio tuning, end-to-end run)


ymohanty:
Component **(a)** of the five-component demo. Ref: [ROADMAP.md §5.1](../blob/main/docs/ROADMAP.md).

Phases 5 and 8 built the pipeline. Two things stop it being demonstrable: Presidio has never been scored against gold data, and the pipeline has never run end to end. All six stages are implemented; **end-to-end integration is unverified.**

**The gold-set work itself is tracked in #15.** It is on the critical path for two other things — the privacy scorecard and the privacy half of the Sarvam comparison (#53) — and finishes in week 1 for that reason.

### Presidio tuning (see #15)
- [ ] Score **per entity**, not aggregate. The release-critical metric is the **false-negative rate** — leaked PII. F1 hides it.
- [ ] Tune recognizer confidence thresholds, context enhancers, and the custom Indian recognizers (mobile / Aadhaar / PAN) against gold.
- [ ] Per-language slices: Odia and romanized Odia, not just English. spaCy NER has no Odia name model.
- [ ] DSI reference: 80.56% any-overlap on a 106-sentence English validation split. **Reference, not threshold**, and not like-for-like with the 85-page set.

### End-to-end run
The `pipeline-core` / `ocr-deepseek` / `categorizer` extras are mutually exclusive, so this is not one pytest.
- [ ] `scripts/e2e_pipeline.sh` — scripted multi-environment run over a fixed sample: document → S3 → artifact DB → exporter → OLTP → materialize → lake → API read-back.
- [ ] Assert row counts at each hop.
- [ ] **Non-PII synthetic variant that runs in CI.** Doubles as the readiness canary: real history queryable, DB result store selected, routing mappings loaded (not silent `fallback`), submit→persist→fetch succeeds.

### ⚠️ Assert the guarantee that actually holds

**"No un-redacted PII reaches the lake" is false and must not be an acceptance criterion.** Two separate reasons:

1. The `complaints` structured columns (`petitioner_name`, `petitioner_mobile`, `petitioner_email`, `address`) are carried faithfully from the dump into both OLTP and the lake. #50's dedup index keys off them.
2. `materialize.py` copies `SELECT *`, so **`complaints.grievance` — the citizen's own account of the problem — is in Parquet verbatim and has never met `pii_tagger`.**

[...]

PR #59, opened 2026-07-28
Phase 13: hold raw OCR back from the lake, add the synthetic E2E canary

ymohanty:
## Summary

Phase 13 (#49), end-to-end run, synthetic half. The real-data variant (`scripts/e2e_pipeline.sh`) needs S3, the model mirrors and the three mutually exclusive extras, and is not in this PR.

Two things: a PII leak into the lake, and the canary that would have caught it.

## What Changed

**Raw OCR was reaching the lake.** `export.py` copies every `PipelinePage` column into OLTP, and `materialize.py` did `COPY (SELECT * FROM oltp.pages)`, so `pages.extracted_text` landed in `pages.parquet` verbatim. That column is un-redacted page text by construction: `pii_tagger` reads it and writes its output to `redacted_text`, leaving the original alone.

- Fixed with `LAKE_COLUMN_DENYLIST` in `olap/materialize.py`. The column stays in OLTP, which is access-controlled; it does not reach Parquet, which is not.
- No schema change, so no Alembic migration.
- Nothing downstream reads it: the summarizer and categorizer take `redacted_text` off the artifact DB, `/history` reads `complaints`, and Phase 17's OCR benchmark reads the artifact DB directly.
- `ExtractionResult.extracted_text` in the frozen serving contract is untouched. Returning raw OCR to the submitter looks deliberate (they are the data subject, and the UI shows extracted vs redacted side by side).

**`tests/test_e2e_synthetic.py`** — artifact DB → OLTP → lake → API, row counts asserted at each hop, with the models replaced by a canned stand-in and everything else on the real code path. Routing is not a model, so it stays real.

The store and history assertions go through `inference.serve.create_live_app` — the function the deployed API actually calls — with only `build_processor` substituted, since that is the piece that loads models. Store selection, history selection and the `Settings` lookup all run for real.

| Guarded | Failure it catches | Asserted through |
|---|---|---|
| Raw page text in Parquet | the leak above, and any regression of it | `materialize` |
| `/history` reads the lake | silently serving `MockHistory`'s fake rows | `create_live_app` |
| DB result store selected | in-memory default, submissions lost on restart | `create_live_app` |
| Routing resolves off master tables | silent `method="fallback"` when the CSVs were never `dvc pull`ed | `MappingRouter` directly |

Mutation-tested by breaking the **production** wiring in `serve.py` (not the tests) and confirming the canary reddens: `MockHistory` instead of `LakeHistory`, always-`InMemoryResultStore`, and always-`DatabaseResultStore`. All fail correctly.

Also pins the silent fallback as behaviour: `_resolve_explicit_oltp_url` treats the built-in SQLite default as "not configured", so an unset `OLTP_DB_URL` gives a healthy API that returns 201 and loses every submission on restart. That is what the demo does if the box's `deploy/.env` omits it.

**Routing is the honest exception.** The router is wired inside `build_processor` (`router=DEFAULT_ROUTER`), which cannot run here, so the synthetic variant covers the loader and not the wire-up. The real-data E2E covers that. The docstring says so rather than implying otherwise.

**Roadmap.** Corrects the stated guarantee. "No un-redacted PII reaches the lake" is false and cannot be an acceptance criterion, because `complaints.grievance` is carried from the dump verbatim and has never met `pii_tagger`. The assertion is about page text; access controls the rest. This matches the correction already written into #49.

## Validation

- [x] `uv run ruff check .` — all checks passed
- [x] `uv run --extra serving --extra pipeline-core pytest` — 280 passed, 7 skipped (rebased onto `main` after #58/#29 merged)
- [x] `uv run --all-groups --extra serving pytest tests/test_e2e_synthetic.py tests/test_materialize.py` — green in the exact CI environment, no heavy extras
- [x] Mutation-tested the guarded properties by breaking production wiring (see above)
- [x] Notebooks have been stripped with `uv run nbstripout --install` or equivalent.
- [x] No proprietary data files are committed directly to Git.
- [x] DVC pointers / `dvc.lock` are updated — lake re-materialized and pushed, see below.

Every fixture in the new test is invented. No citizen text enters the test suite, which is what makes it runnable in CI.

## Review Expectations

- Keep this PR small enough to be reviewed by a human in a short sitting.
- Split unrelated changes into separate PRs before requesting review.
- Two-person review is required.
- Yashaswi is a mandatory reviewer.
- Add one additional reviewer who can assess the code, analysis, or domain impact.

The materialization change touches what leaves the trust boundary, so this wants the security-sensitive review tier.

## Data And Delivery Notes

- [ ] This PR does not modify data.
- [x] This PR ingests or versions data through DVC.
- [ ] This PR updates generated exhibits.

Notes:

### DVC artifact re-materialized (Codex P1, resolved)

The code fix alone left the leak in circulation: `dvc.lock` still pointed at the pre-fix `pages.parquet`, so `dvc pull` on a fresh checkout kept receiving the Parquet containing `extracted_text`.

Re-materialized and pushed by the maintainer. The lock now records:

| | Before | After |
|---|---|---|
| dep `janasunani/olap/materialize.py` | `ea29e1c5…` (3,356 B) | `9bb2ff8d…` (5,273 B) |
| out `data/interim/pages.parquet` | `816be322…` (10,361 B) | `4ad7a178…` (3,925 B) |

The recorded dep hash matches the file carrying `LAKE_COLUMN_DENYLIST` byte for byte, so the recorded output is the one that code produced — not a re-run of the old stage. The other three tables are unchanged, as expected, since only `pages` has a denied column.

**On the pre-fix blob**, which stays in the remote under its old hash: no retraction needed. The DVC remote is IAM-scoped, and everyone who can reach it already has access to the raw dump — 1.37M grievances with names, mobiles and addresses. A few OCR'd pages from a 2-file sample add nothing to that. The exposure this PR closes was `dvc pull` at HEAD handing raw page text to anyone entitled to the *lake*, which is a wider group than the one entitled to the dump.

- Rebased onto `main` after #58 and #29 merged. Suite grew 230 → 280 as a result (#29's deploy tests).
- Roadmap now edits the rewritten version #29 brought to `main`. Two corrections there: §3.2's `pages.extracted_text` row claimed it "never leaves the pipeline artifact DB" (the exporter copies it to OLTP), and §5.1's E2E bullet asserted "no un-redacted PII reaches the lake", contradicting the precise guarantee stated in §3.2 of the same document.

### DIFF
--- a/docs/ROADMAP.md
+++ b/docs/ROADMAP.md
@@ -227,9 +227,9 @@ characters). Phase 15 S4 then reads that field over the full corpus by design.
 |---|---|---|
 | `complaints` structured columns, OLTP **and** lake | **Yes** | `petitioner_name`, `petitioner_mobile`, `petitioner_email`, `address`. Faithful to the dump, by design |
 | **`complaints.grievance`, OLTP and lake** | **Yes, raw prose** | Citizen-authored, never passed through `pii_tagger`. Materialized verbatim by `SELECT *` |
-| `pages.extracted_text` (raw OCR) | **Yes** | Never leaves the pipeline artifact DB |
-| `pages.redacted_text` | No, by construction | Typed tokens. This is what the exporter carries onward |
-| Parquet lake, *pipeline-derived* text fields | No | Only redacted page text is exported by the pipeline |
+| `pages.extracted_text` (raw OCR) | **Yes** | Artifact DB **and OLTP** — the exporter copies every column. Held back from the lake by `LAKE_COLUMN_DENYLIST` (`olap/materialize.py`); access-controlled, not redacted |
+| `pages.redacted_text` | No, by construction | Typed tokens. This is what reaches the lake |
+| Parquet lake, *pipeline-derived* text fields | No | Only redacted page text. Enforced by the denylist above and asserted in `tests/test_e2e_synthetic.py`, not merely intended |
 | Dedup index, MinHash signatures | **Yes, derived** | Salted contact hashes, plus signatures over raw `grievance`. A hash is not anonymous |
 | Embeddings over `grievance` (S4) | **Yes, derived** | A vector is not anonymous merely because it is unreadable |
 | API responses, summaries | No | Built only from redacted text |
@@ -442,12 +442,28 @@ pytest.
 - `scripts/e2e_pipeline.sh`: a scripted multi-environment run over a fixed sample.
   Document → S3 → artifact DB → exporter → OLTP → materialize → lake → API
   read-back.
-- Asserts row counts at each hop, and that no un-redacted PII reaches the lake or
-  any API response.
+- Asserts row counts at each hop, and that no un-redacted *page* text reaches the
+  lake. **Not** "no un-redacted PII reaches the lake": §3.2 explains why that is
+  false and cannot be an acceptance criterion.
 - A **non-PII synthetic variant** that runs in CI. This doubles as the readiness
   canary: it proves real history is queryable, the DB result store is selected,
   routing mappings loaded rather than silently `fallback`, and submit→persist→fetch
   succeeds.
+
+✅ **The synthetic variant is built** (`tests/test_e2e_synthetic.py`, PR #59).
+Artifact DB → OLTP → lake → API over invented fixtures, row counts asserted at
+each hop, with only the models replaced by a canned stand-in. History and store
+selection are asserted through `inference.serve.create_live_app`, so a
+deployment that silently keeps `MockHistory` or `InMemoryResultStore` fails the
+canary. Two caveats it records honestly: routing is exercised via `MappingRouter`
+directly, because the router is wired inside `build_processor`, which needs the
+models; and the pipeline's own stages are not run. Both belong to the real-data
+variant, which is still outstanding.
+
+Building it found the leak §3.2 now describes: `materialize.py` was copying
+`pages.extracted_text` into Parquet verbatim. The lake was re-materialized and
+`dvc.lock` updated, because the code fix alone left the pre-fix artifact in
+circulation through `dvc pull`.
 
 **Exit criteria.** Per-entity, per-language PII scorecard published. E2E green on
 real data and synthetic in CI. DSI baselines reproduced or the gap explained.
--- a/dvc.lock
+++ b/dvc.lock
@@ -9,8 +9,8 @@ stages:
       size: 7071113216
     - path: janasunani/olap/materialize.py
       hash: md5
-      md5: ea29e1c583a082cf9bc669f98396f72b
-      size: 3356
+      md5: 9bb2ff8dbbbcc3c9502f5d080c0fbc11
+      size: 5273
     outs:
     - path: data/interim/action_history.parquet
       hash: md5
@@ -26,8 +26,8 @@ stages:
       size: 198
     - path: data/interim/pages.parquet
       hash: md5
-      md5: 816be322183d304d67c042f27ae55d9a
-      size: 10361
+      md5: 4ad7a17893cb22a396f572b899e68c25
+      size: 3925
   pipeline-sample:
     cmd: uv run --extra pipeline-core janasunani-pipeline run --input 
       data/raw/documents-sample --db data/processed/pipeline-sample.sqlite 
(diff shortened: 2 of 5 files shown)


ymohanty:
Both findings confirmed against the code. Thanks — the P2 one was a real overclaim on my part.

## P2 (live wiring) — fixed in 5244a39

Correct: the tests injected `LakeHistory` and `DatabaseResultStore` into `create_app`, which proved the components work and said nothing about whether production selects them. My mutation check was mutating the *tests*, which is why it looked green.

Both assertions now go through `inference.serve.create_live_app` with only `build_processor` substituted, since that is the one piece that loads models. Store selection, history selection and the `Settings` lookup run for real. Re-ran the mutation check against **production wiring** in `serve.py`:

| Mutation to `serve.py` | Canary |
|---|---|
| `history=MockHistory()` instead of `LakeHistory()` | FAIL ✅ |
| always `InMemoryResultStore()` | FAIL ✅ |
| always `DatabaseResultStore(...)` | FAIL ✅ |

Also pinned the silent fallback as behaviour: `_resolve_explicit_oltp_url` treats the built-in SQLite default as "not configured", so an unset `OLTP_DB_URL` yields a healthy API that returns 201 and loses every submission on restart. That is what the demo does if the box's `deploy/.env` omits it.

One scope correction rather than a fix: **routing stays a direct `MappingRouter` test**, because the router is wired inside `build_processor` (`router=DEFAULT_ROUTER`), which cannot run in this environment. The synthetic variant covers the loader, not the wire-up. The module docstring and the PR body now say that instead of implying otherwise; the real-data E2E covers the rest.

## P1 (DVC artifact) — confirmed, and it blocks merge

Verified in `dvc.lock`: dep `janasunani/olap/materialize.py` at the pre-fix hash `ea29e1c5…`, out `data/interim/pages.parquet` at pre-fix `816be322…`. So `dvc pull` still serves the Parquet containing `extracted_text`. The code fix ships the filter while the remote keeps serving the leak — agreed, and my PR body understated this as "stale files".

I cannot fix it here: it requires running `materialize` against the 7 GB local SQLite OLTP under `data/`, which I am not permitted to access ([AGENTS.md §Data Access Restriction](../blob/main/AGENTS.md)), plus S3 credentials for `dvc push`. Handed to the maintainer with the exact steps, and the PR checkbox corrected to unchecked.

**One thing beyond the finding:** re-materializing fixes what `dvc pull` serves at HEAD, but the pre-fix blob stays in the DVC remote under its old hash, reachable from any older commit. That is a retention decision, not a lockfile one.

Suite: 230 passed, 7 skipped. `ruff` clean.

ymohanty:
P1 resolved, and the branch is rebased onto `main` now that #58 and #29 have merged.

## P1 — the DVC artifact

Correct that the code fix alone left the leak in circulation. `dvc.lock` still pointed at the pre-fix `pages.parquet`, so `dvc pull` on a fresh checkout kept receiving the Parquet containing `extracted_text`.

Re-materialized and pushed:

| | Before | After |
|---|---|---|
| dep `janasunani/olap/materialize.py` | `ea29e1c5…` (3,356 B) | `9bb2ff8d…` (5,273 B) |
| out `data/interim/pages.parquet` | `816be322…` (10,361 B) | `4ad7a178…` (3,925 B) |

The recorded dep hash matches the file carrying `LAKE_COLUMN_DENYLIST` byte for byte, so the recorded output is what that code produced rather than a re-run of the old stage. `complaints`, `action_history` and `documents` are unchanged, as expected — only `pages` has a denied column.

**One thing the finding did not cover, flagged rather than fixed:** the pre-fix blob is still in the DVC remote under its old hash and reachable from any earlier commit. Re-materializing fixes what `dvc pull` serves at HEAD; it does not retract what is already stored. That is a retention decision for the maintainer, not a lockfile one.

## Rebase

`main` moved 55 commits (#58, then #29). Rebased rather than merged, so the three commits stay reviewable. One conflict, in `docs/ROADMAP.md`, because #29 replaced the 600-line roadmap with the rewritten 1,901-line one. Resolved by taking the new version and re-applying the correction where it actually belongs:

- **§3.2** listed `pages.extracted_text` as "Never leaves the pipeline artifact DB". It does — `export.py` copies every column into OLTP. Corrected to say it reaches OLTP and is held back from the lake by the denylist, controlled by access rather than redaction.
- **§5.1** asserted the E2E should check "no un-redacted PII reaches the lake", which contradicts the precise guarantee §3.2 states three hundred lines earlier. Corrected to the page-text guarantee, with a pointer to why the broader claim is false.

## Validation on the new base

- `uv run --extra serving --extra pipeline-core pytest` — 280 passed, 7 skipped (was 230; #29 brought the deploy tests)
- `uv run --all-groups --extra serving pytest tests/test_e2e_synthetic.py tests/test_materialize.py` — 13 passed, 1 skipped in the exact CI environment

@codex review

Issue #30, opened 2026-07-21
Deploy: first live bring-up of the demo stack on the CPU box


ymohanty:
Maintainer-owned, one-time steps to take the demo stack live on the always-on CPU box (`i-0ef24e15a80ba7128`, EIP `52.66.116.80`). The CI → GHCR → box deploy automation is built and reviewed on `deploy/cpu-box`, but the deploy has **never actually run** — these steps can only be done by the maintainer (they touch AWS, GHCR, and the box holding production PII).

Full runbook: [docs/DEPLOY.md §4](../blob/main/docs/DEPLOY.md) ("One-time box setup" + "Routine flow").

### Checklist
- [ ] **Box setup**: `docker login ghcr.io` with a `read:packages` PAT; `cp .env.example .env` + fill `POSTGRES_PASSWORD` / `SITE_ADDRESS=52-66-116-80.nip.io`; `cp proxy.env.example proxy.env` + set a real `DEMO_PASSWORD_HASH` (via `caddy hash-password`). Both files `chmod 600`.
- [ ] **Scoped `dvc pull`** on the box: categorizer, page-type ViT, and `janasunani-mappings` (do NOT bare `dvc pull`). Without the mappings, routing degrades to `method:"fallback"`.
- [ ] **CI IAM role**: `terraform apply` of `deploy/terraform/ci.tf` (OIDC provider + `ci_deploy` role). `terraform plan` first — abort if anything on `aws_instance.cpu_box`/`aws_security_group.cpu_box` shows destroy/replace.
- [ ] **GitHub `box-deploy` environment**: create it; add yourself as a required reviewer; set the six secrets/vars (`BOX_SSH_KEY`, `BOX_SSH_KNOWN_HOSTS`, `BOX_HOST`, `CI_DEPLOY_ROLE_ARN`, `BOX_SG_ID` on the **environment**; `DPIC_GITHUB_SSH_KEY` stays repo-level).
- [ ] **First deploy**: Actions → "Deploy demo" → Run workflow (empty `image_tag` builds from default branch). Confirms the `linux/amd64` api image builds for real (can't be validated on an arm64 dev Mac).
- [ ] **On-box browser E2E** (do once after the first deploy): submit a grievance → real pipeline output renders and persists to `live_grievances` → `/history` shows it → `basic_auth` actually gates the site.

### Notes
- Deferred until we're ready to go live for the demo. No code work remains here — it's ops.
[...]

PR #62, opened 2026-07-28
Deploy: CPU-only Torch for the api image (#48)

### LINKED ISSUE #48: Deploy: CPU-only Torch image for the always-on box
The api image (~8-12 GB) bundles CUDA Torch even though the always-on CPU box is CPU-only and DeepSeek is deliberately excluded. Ref: [ROADMAP.md §4 Phase 12](../blob/main/docs/ROADMAP.md) + CTO deploy/cpu-box addendum.

### Why
A CPU-only Torch build materially cuts CI time, pull time, disk pressure, rollback time, and the chance a deploy exhausts shared disk and affects the production database.

### Checklist
- [ ] Pin the CPU-only Torch wheel (extra-index) for the `demo`/api image build.
- [ ] Verify the live processor still loads (page-type ViT, MuRIL, BART, Presidio) on CPU.
- [ ] Confirm image size drop + rollback-time improvement.

ymohanty:
## Summary

Closes #48. The always-on box is CPU-only and hosts the production Postgres, but the api image was shipping the full CUDA runtime.

On `linux/amd64` the `demo` extra pulled `cuda-bindings`, `cuda-toolkit` (cudart, cufft, cufile, cupti, curand, cusolver, cusparse, nvjitlink, nvrtc, nvtx), `nvidia-cublas`, `nvidia-cudnn-cu13`, `nvidia-cusparselt-cu13`, `nvidia-nccl-cu13`, `nvidia-nvshmem-cu13` and `triton` — behind a torch wheel that is 532 MB on its own.

None of it was ever reachable. DeepSeek OCR is deliberately excluded from `demo`, so nothing in that image can use a GPU.

## What Changed

`demo` now resolves torch from PyTorch's CPU-only index.

| | Before | After |
|---|---|---|
| torch wheel (linux x86_64) | 532.3 MB | 192.3 MB |
| torch's CUDA payload | ~2,196 MB across 17 packages | none |
| **Total wheels removed** | | **~2.5 GB** (more once unpacked) |

**One CUDA wheel survives and the torch source cannot remove it.** `xgboost` declares `nvidia-nccl-cu12` unconditionally on linux, and `demo` pulls xgboost through `pipeline-core`, so ~300 MB remains. Verified against the real lock:

```
$ uv export --frozen --extra demo --no-dev --no-hashes --no-emit-project | grep -iE '^(nvidia|torch|xgboost)'
nvidia-nccl-cu12==2.28.3 ; sys_platform == 'linux'
torch==2.12.1 ; sys_platform == 'darwin'
torch==2.12.1+cpu ; sys_platform != 'darwin'
xgboost==3.3.0
```

`build_processor` never loads the format classifier, xgboost's only consumer, so it is dead weight in this image — filed as #63. Removing it changes what `demo` includes, a different decision from where torch comes from, so it is not in this PR.

**Scoped to `demo` alone, deliberately.** `categorizer` and `pipeline-core` keep CUDA wheels, because the MuRIL embedding pass is a corpus-scale GPU batch job on the GPU box (ROADMAP §5.3 S4), and `ocr-deepseek` is GPU-only by nature. Pinning those would have removed GPU acceleration from the backfills to save space in an image they are not built into. There is a test asserting they stay unpinned.

Two implementation notes worth flagging in review:

- `torch` is listed **directly** on `demo`, not only inherited via `janasunani[pipeline-core]`. uv binds a source's `extra =` marker only to packages declared on that extra, and errors out otherwise.
- It is pinned to the already-locked `2.12.1`. Without the pin, adding the direct requirement re-resolves and drags torch to 2.13.0 and **torchvision to 0.28.0 in the GPU path**, which nothing in this repo can test. With it, the lock diff is exactly one added version.

Also updated the four places describing the image as CUDA-bearing: `deploy/api.Dockerfile`, `.github/workflows/deploy.yml`, `docs/DEPLOY.md` (×2), `docs/ROADMAP.md` §Phase 12. The `deploy.sh` ~20 GiB free-space guard is **left where it is** — the point is headroom for prod Postgres and the lake, not a tight fit around the image.

## Validation

- [x] `uv run ruff check .` — all checks passed
- [x] `uv run --extra serving --extra pipeline-core pytest` — 285 passed, 7 skipped (5 new)
- [x] `uv lock --check` — consistent, so the Dockerfile's `uv sync --locked` still works
- [x] `uv sync --extra demo` + import smoke — torch 2.12.1, transformers 4.57.6, loads clean
- [x] Verified against a **real resolution** (`uv export --frozen --extra demo`, markers evaluated for linux) that torch is `2.12.1+cpu` and `nvidia-nccl-cu12` is the only remaining CUDA package
- [x] Verified the GPU path is untouched: `torchvision` still 0.27.1, the PyPI torch entry still carries its 7 CUDA deps
- [x] Mutation-tested every new assertion (removed the source and re-locked; made the index non-explicit; wrongly pinned `categorizer`) — all fail correctly
- [x] Notebooks have been stripped with `uv run nbstripout --install` or equivalent.
- [x] No proprietary data files are committed directly to Git.
- [x] DVC pointers / `dvc.lock` are updated when approved data or pipeline outputs changed.

### ⚠️ What is NOT verified here

**The image-size drop itself.** On darwin the `demo` extra still resolves the ordinary PyPI wheel — the CPU source is markered `sys_platform != 'darwin'` — so the win is invisible on an arm64 dev Mac by construction. The `linux/amd64` build is CI's job.

First real signal, in the build log:

- `torch-2.12.1+cpu-...-manylinux_2_28_x86_64.whl` downloaded
- **exactly one** `nvidia-*` download (`nvidia-nccl-cu12`, #63). Any other `nvidia-*`/`cuda-*` wheel means the CPU source did not take effect
- the pushed image materially smaller than the previous ~8–12 GB

Also unverified: that the live processor (page-type ViT, MuRIL, BART, Presidio) loads on CPU torch on amd64. It loads fine on darwin CPU torch here, which is the same code path minus the platform, but the box is the real test. That is #30's on-box browser E2E.

## Review Expectations

- Keep this PR small enough to be reviewed by a human in a short sitting.
- Split unrelated changes into separate PRs before requesting review.
- Two-person review is required.
- Yashaswi is a mandatory reviewer.
- Add one additional reviewer who can assess the code, analysis, or domain impact.

The dependency-resolution change is the substance; the rest is comments and docs catching up. `uv.lock` is +95/-14 and mostly marker churn.

## Data And Delivery Notes

- [x] This PR does not modify data.
- [ ] This PR ingests or versions data through DVC.
- [ ] This PR updates generated exhibits.

Notes:

- Lands before the first deploy on purpose. Doing the first-ever `workflow_dispatch` with the oversized image is the version of #30 with the most downside, since a disk-full on that box risks prod Postgres, the lake and the nightly `pg_dump` target, not just the deploy.
- The GHCR registry cache will miss on the first build after this, since the dependency layer changes. Expect one slow build.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -67,8 +67,8 @@ jobs:
           # commits. The SHA tag is always both-or-neither (this job's push
           # either succeeds or the whole job — and thus the deploy — fails).
           tags: ${{ env.API_IMAGE }}:${{ github.sha }}
-          # Registry cache, not GHA cache: the `demo` extra resolves CUDA
-          # torch (~8-12 GB image) — well past the GHA cache size budget.
+          # Registry cache, not GHA cache: even on CPU-only torch the `demo`
+          # extra's layers run past the GHA cache size budget.
           # (This `buildcache` tag is BuildKit's layer cache, unrelated to
           # image tagging above — it's never pulled/deployed.)
           cache-from: type=registry,ref=${{ env.API_IMAGE }}:buildcache
--- a/deploy/api.Dockerfile
+++ b/deploy/api.Dockerfile
@@ -32,8 +32,12 @@ RUN mkdir -p -m 0700 /root/.ssh && ssh-keyscan github.com >> /root/.ssh/known_ho
 COPY pyproject.toml uv.lock ./
 # Dependency-only sync first (no project code yet) so this layer caches across
 # source-only changes; the registry cache (build-push-action cache-from/to
-# type=registry) makes this fast even on a cold CI runner given the ~8-12 GB
-# CUDA-torch resolve for `demo`.
+# type=registry) keeps this fast on a cold CI runner.
+#
+# `demo` resolves torch from PyTorch's CPU-only index (see [tool.uv.sources] in
+# pyproject.toml) — this box has no GPU. That drops ~2.5 GB of wheels versus the
+# default CUDA build: the nvidia-*/cuda-toolkit/triton payload entirely, plus
+# the torch wheel itself going 532 MB -> 192 MB.
 RUN --mount=type=ssh uv sync --locked --extra demo --no-dev --no-install-project
 COPY README.md alembic.ini ./
 COPY janasunani ./janasunani
--- a/deploy/api.Dockerfile
+++ b/deploy/api.Dockerfile
@@ -36,8 +36,13 @@ COPY pyproject.toml uv.lock ./
 #
 # `demo` resolves torch from PyTorch's CPU-only index (see [tool.uv.sources] in
 # pyproject.toml) — this box has no GPU. That drops ~2.5 GB of wheels versus the
-# default CUDA build: the nvidia-*/cuda-toolkit/triton payload entirely, plus
-# the torch wheel itself going 532 MB -> 192 MB.
+# default CUDA build: torch's whole cuda-toolkit/cuda-bindings/nvidia-*/triton
+# payload, plus the torch wheel itself going 532 MB -> 192 MB.
+#
+# One CUDA wheel survives: xgboost declares nvidia-nccl-cu12 unconditionally on
+# linux, and `demo` pulls xgboost via pipeline-core. The live processor never
+# loads the format classifier, so it is ~300 MB of dead weight here — tracked
+# separately, not fixed by the torch source.
 RUN --mount=type=ssh uv sync --locked --extra demo --no-dev --no-install-project
 COPY README.md alembic.ini ./
 COPY janasunani ./janasunani
--- a/docs/DEPLOY.md
+++ b/docs/DEPLOY.md
@@ -389,8 +389,12 @@ policy itself prevents it.
   Dockerfile carefully; the first real signal is the CI build log / GHCR push.
   This is also where the CPU-torch switch (#48) gets confirmed: on darwin the
   `demo` extra still resolves the ordinary PyPI wheel, so the image-size drop
-  is not observable locally. Check the build log for
-  `torch-2.12.1+cpu-...manylinux_2_28_x86_64.whl` and no `nvidia-*` downloads.
+  is not observable locally. In the build log, expect
+  `torch-2.12.1+cpu-...manylinux_2_28_x86_64.whl`, and **exactly one**
+  `nvidia-*` download — `nvidia-nccl-cu12`, which xgboost declares
+  unconditionally on linux (~300 MB, tracked separately). Any other `nvidia-*`
+  or `cuda-*` wheel means the CPU source did not take effect and the image is
+  still carrying the CUDA runtime.
 - **On-box browser E2E** — submit a grievance → real pipeline output renders
   and persists to `live_grievances` → `/history` shows it → `basic_auth`
   actually gates access. Do this once after the first automated deploy.
--- a/docs/DEPLOY.md
+++ b/docs/DEPLOY.md
@@ -372,7 +372,10 @@ policy itself prevents it.
   Parquet lake, the HF cache, and the nightly `pg_dump` target — a
   disk-full here risks all of those, not just the deploy. `deploy.sh`
   refuses to even start pulling images if free space drops below ~20 GiB
-  (a few multiples of one ~8-12 GB api image), and after a successful
+  (deliberately several multiples of one api image; the threshold is kept
+  where it was after the CPU-torch switch shrank the image, since the point
+  is headroom for prod Postgres and the lake, not a tight fit), and after a
+  successful
   deploy prunes every SHA-tagged `janasunani-api`/`janasunani-frontend`
   image except the current and previous (the rollback target) — plus the
   usual `docker image prune -f` for dangling layers. If disk pressure still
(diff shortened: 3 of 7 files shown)


ymohanty:
CI is green, but the two long jobs need a note so they are not misread.

`ruff` took 5m47s against 23–43s on every other recent run, and `test-and-validate-pipeline` 4m33s against ~1m15s. That is cache invalidation, not a regression:

```
Resolved  349 packages in 443ms     <- indexes are fine
Prepared  273 packages in 5m 27s    <- cold cache, wheels re-downloaded
Installed 273 packages in 856ms
```

`setup-uv@v5` keys its cache on the `uv.lock` hash. This PR changes the lock, so the first run on the new key downloads everything. Resolution at 443ms confirms the new `pytorch-cpu` index is not being probed on syncs that do not install torch, which is what `explicit = true` is for.

Self-healing: the first run after merge repopulates the cache and subsequent runs return to ~30s. Same one-time cost as the GHCR layer-cache miss already flagged in the PR body.

ymohanty:
Confirmed and fixed in 4ba0476. This was a real defect in the test, and it made the documentation wrong too.

## The finding is exactly right

My check only treated a CUDA edge as reaching `demo` when that edge's own marker named the extra. It cannot see a package pulled by an *unconditional* edge from a parent that `demo` selects — which is precisely this shape:

```
demo -> pipeline-core -> xgboost -> nvidia-nccl-cu12 ; sys_platform == 'linux'
```

`xgboost`'s edge names no extra, so my filter skipped it. The test passed while the api image still downloaded the NCCL wheel, and `docs/DEPLOY.md` told the operator to expect no `nvidia-*` downloads at all.

## Fixed by resolving instead of inferring

Replaced the per-edge inspection with an actual resolution, as suggested — `uv export --frozen --extra demo` against the real lock, markers evaluated for the box's platform:

```
nvidia-nccl-cu12==2.28.3 ; sys_platform == 'linux'
torch==2.12.1 ; sys_platform == 'darwin'
torch==2.12.1+cpu ; sys_platform != 'darwin'
xgboost==3.3.0
```

The new test asserts torch is the `+cpu` build and that `nvidia-nccl-cu12` is the **only** remaining CUDA package — named explicitly rather than filtered out, so the surface cannot grow back silently. Mutation-tested by removing the CPU source and re-locking; both assertions fail.

## Corrected the claims

Three places said the CUDA payload goes entirely: `deploy/api.Dockerfile`, `docs/DEPLOY.md`, `docs/ROADMAP.md`. DEPLOY.md now tells the operator to expect **exactly one** `nvidia-*` wheel and to treat any other as the CPU source having failed to take effect — which is a more useful check than the one I had written, since it now discriminates.

## The surviving NCCL: #63

Worth adding beyond the finding — `xgboost` is only used by the format classifier, and `build_processor` never loads it. So the api image ships xgboost and ~300 MB of NCCL beneath it for a stage that image never runs. Filed as #63 with two fix directions. Not folded in here because removing it changes what `demo` includes, which is a different decision from where torch comes from.

Suite 285 passed / 7 skipped, `ruff` clean, `uv lock --check` consistent.

@codex review

ymohanty:
Confirmed and fixed in 5c8120c. This was a real security regression and entirely my error — thank you for catching it.

## What happened

`f18c1b4` shipped a corrupted `uv.lock`. Verified against `main`:

| | main | f18c1b4 | now |
|---|---|---|---|
| certifi | 2026.6.17 | **2022.12.7** | 2026.6.17 |
| requests | 2.34.2 | **2.28.1** | 2.34.2 |
| urllib3 | 2.7.0 | **1.26.13** | 2.7.0 |
| packages | 350 | 349 | 352 |

Plus a dozen unrelated downgrades, none of which had anything to do with the torch source.

## Cause — the exact hazard `explicit = true` exists to prevent

`download.pytorch.org` is not a torch-only mirror; it also serves old copies of certifi, requests and urllib3. While **mutation-testing the `explicit` assertion**, one mutation set it to `false`, and `uv run pytest` re-locked against the mutated pyproject. uv then treated the PyTorch index as general and resolved the TLS stack from it. My cleanup restored `pyproject.toml` but not `uv.lock`, so the poisoned lock got committed.

The check that would have caught it was in my own output and I read past it: `uv lock --check` reported **349 packages** where the correct lock has 352. I saw the number and did not compare it to the 352 I had seen twenty minutes earlier.

## Fix

Lock regenerated from `main`'s. The diff is now what it should always have been:

```
2 +name = "torch"
1 +version = "2.12.1+cpu"
1 +version = "2.12.1"
```

Nothing else moves.

## Guard added

`test_only_torch_is_sourced_from_the_pytorch_index` asserts no package except torch resolves from `download.pytorch.org`. Deliberately asserted against **the lock, not the pyproject flag** — the flag being correct does not prove the lock was generated while it was, which is precisely how this slipped through. Verified by running it against the actual corrupted lock from `f18c1b4` rather than a synthetic one: it fails, naming the leaked packages.

Suite 286 passed / 7 skipped, `ruff` clean, `uv lock --check` back to 352.

@codex review

ymohanty:
@codex review

Issue #72, opened 2026-08-05
Duplicate recall against the ~34,700 officer-confirmed duplicates


ymohanty:
Parent: #50. The headline number for component (b), and DELIVERY.md Table 2 row 3.

Two discard-reason template families give roughly **34,700 officer-confirmed
duplicates** ("case already taken up" 19,904 + "duplicate copy" 14,767).

**These are the baseline, not the deliverable.** They are exactly the duplicates the
existing manual process already caught, and they are queryable today with a
`CASE WHEN`. The capability claim is the **increment**: duplicates MinHash finds that
carry no such remark.

### Done when
- [ ] Recall reported against the confirmed set
- [ ] The increment reported **separately**, as the actual claim
- [ ] Prevalence by district, category, mode, year over the slice
- [ ] Duplicate-adjusted workload: how much of the backlog is one problem arriving more than once

⚠️ The confirmed set is not ground truth either. It inherits whatever bias the old
process had. Check discard-rate variance by office before drawing conclusions, or the
finding is about office identity rather than content.

ymohanty:
## Baseline measured — and the first measurement was wrong

Ran the finding over the full lake. Posting both numbers because the first one was published and needs retracting.

### Corrected result

| | Count |
|---|---|
| **Officer-confirmed duplicates** | **39,937** |
| — `taken_up` family | 21,513 |
| — `duplicate_copy` family | 18,424 |
| Resolved complaints carrying a closing remark | 1,209,144 |
| **Officer-confirmed share** | **3.3%** |

ROADMAP reference was 34,671 (19,904 + 14,767); delta **+5,266**.

### What was wrong

The first version of the mart matched by equality:

```sql
WHEN closing_remark IN ('case already taken up', 'case taken up earlier', ...) THEN 'taken_up'
```

The templates are not stored in that form. Measured over the lake:

| Stored template | Count |
|---|---|
| `case already taken up for examination` | 11,736 |
| `case taken up earlier hence closed` | 9,289 |

Equality matched **8** rows where ~21,500 exist. The published finding read **18,432** total (1.5% share) and presented the shortfall against the ROADMAP as a property of the data — "delta −16,239". It was a predicate bug.

Fixed in #173 (`8e38c08`) by anchoring on the prefix.

### Why the corrected number lands *above* the ROADMAP reference

[...]

Issue #73, opened 2026-08-05
Duplicate and campaign treatment on the result screen


ymohanty:
Parent: #50. Advisory triage banner on the result screen:

- "possible duplicate of ticket NNN", with a link
- "part of a campaign, N related filings"
- "flagged low-signal, review"

All advisory, **none blocking**. ROADMAP §5.2: never auto-reject. A false positive is
a citizen's grievance discarded, which is a different class of harm from a
mis-categorization.

Builds against the frozen `serving/schemas.py` contract with mocked responses, so it
does not wait on the backfill. Suitable for `frontend-dpic`.

Note the spam banner ships only if spam scoring does; that is cut to Post-demo, so the
default is the duplicate and campaign treatments alone.

ymohanty:
## Verified the four advisory signals end to end

### UI — all four render, none block

`TriageBanner.tsx` covers every state:

- **Duplicate** — `possible duplicate` badge, "Possible duplicate of ticket NNN" with a `Link` to `/history?q=<ticket>`, and "Review both filings before taking any action." The linked ticket #109 asks for is there.
- **Campaign** — `collective grievance` badge (positive tone, not negative), "Part of a campaign — N related filings … This is a collective grievance, not spam; each filing remains visible for review." Framed exactly as the issue requires.
- **Spam** — `low-signal review` badge, message, `Reason code: <code>`, and "This advisory does not reject the grievance."
- **Not-found states are shown, not hidden** — separate branches for `not_indexed`, `unavailable`, `abstained` and `no_match`, each carrying its reason, with "This is not a finding that there are no related filings."
- **Routing** — destination and confidence come from the existing `RoutingResult` on `main` (crosswalk, #33).

### Backend — spam and routing live, duplicate and campaign unwired

The production default is `UnwiredTriageProvider`:

```python
def assess(self, *, redacted_text, district, submitted_on) -> TriageResult:
    # Duplicate matching remains unwired.
    del district, submitted_on
    return TriageResult(spam=low_signal_advisory(redacted_text))
```

- **Spam — satisfied.** `low_signal_advisory` always returns `decision="abstained"`, with `ocr_repetition_collapse_unvalidated` when the OCR repetition guard fires and `live_review_disabled_pending_redacted_adjudication` otherwise. #109 asks for the abstention to be *shown rather than hidden*, and it is — with its reason code and evidence. Deliberate: no rule may fire before a human-adjudicated validation release.
[...]

PR #135, opened 2026-08-06
feat(pipeline): dedup index backfill runner (#71)

### LINKED ISSUE #71: Run the redaction pass and build the dedup index over the slice
Parent: #50. The calendar item, not a sprint item. ROADMAP §5.6 B: "write the stage
early, kick the backfill, work on something else while it runs. Starting either late
in week 3 is a plan to fail."

Strict order:
```
janasunani-redact-grievance ──► dedup index build ──► duplicate scoring
```

### Done when
- [ ] Redaction pass complete over the chosen slice, **before** anything indexes it
- [ ] Dedup index built from `grievance_redacted`, never from the raw column
- [ ] Row counts reconciled against the slice definition
- [ ] Index artifacts marked `dpic-infra`: DPIC machines only, never an `authorized-external` route

⚠️ Redaction lowers exposure; it does not declassify what is derived. Signatures
built from citizen prose inherit its classification, because distinctive phrasing
re-identifies where a phone number no longer does (ROADMAP §3.2).

Blocked by the slice decision and by the two authoring issues. One box, one
Postgres, one lake: this cannot run concurrently with another backfill.

ymohanty:
Fixes #71

## What this is

`janasunani/pipeline/dedup.py` (merged earlier) provides the MinHash/LSH
primitives and is deliberately stdlib-only. Nothing walked a slice, computed
signatures, blocked candidates, verified them and persisted a result — that's
`janasunani/pipeline/dedup_index.py`, one stage after
`janasunani-redact-grievance` in the backfill order (ROADMAP §5.2):

```
janasunani-redact-grievance -> dedup index build (this, #71) -> spam_duplicate scoring
```

It mirrors `redact_grievance.py`'s shape on purpose: argparse entrypoint,
`def main() -> None` wrapping `asyncio.run`, its own `create_async_engine`,
`--district`/`--year` with no defaults, batched and resumable via a pending
predicate (not an offset).

Run: `uv run --extra pipeline-core janasunani-dedup-index --district Sambalpur --year 2024`

## Built from `grievance_redactions.grievance_redacted` only

The pending query joins `complaints` only for structured columns (district,
`created_on`, `petitioner_mobile`/`email`) — it never selects
`complaints.grievance`. Two tests pin this directly rather than by
inspection: a pair with identical raw text but different redacted text does
**not** group; a pair with completely different raw text but identical
redacted text **does** group. Either one fails if the source column changes.

## Blocking strategy

A single district-year slice runs to tens of thousands of rows (55,544 for
Sambalpur 2024 — ~1.5B unordered pairs unblocked). Each signature gets a
`block_key` of `district:script:window_index`:

- **`script`** — `"odia"` if the redacted text contains any Odia-script
  codepoint, else `"latin"`. Odia-script and romanized-Odia filings are
  never bucketed together and never compared, on top of the fact that
  character shingles across the two scripts already share ~zero n-grams by
  construction. Cross-script recall is explicitly unsupported (Phase 17
  closes it via transliteration); this is stated in the module docstring
  and pinned by tests.
- **`window_index`** — `(created_on - Jan 1 of --year).days // --window-days`
  (30 days by default), `None`/"undated" for a missing `created_on`.

LSH candidate generation only buckets within one block; a candidate pair is
then verified with exact `jaccard_similarity` before it's allowed to union
(dedup.py's own contract — an LSH band collision is a candidate, not a
confirmed duplicate).

**Identity-key matching (same-citizen resubmission) is deliberately
unblocked** — a resubmission can land months later and a phone number
doesn't carry a script or a window. Salted `identity_key_mobile`/
`identity_key_email` hashes (never derived from the redacted text — a
separate path per `dedup.py`'s contract) are grouped by equality across the
whole slice and unioned in alongside the verified text pairs.

## Persistence: two tables, one alembic revision

`e59fb4410dd6` (head `b7c1e94af203` -> `e59fb4410dd6`):

- **`dedup_signatures`** — one row per ticket: the 128-int MinHash signature
  (JSON, portable across SQLite/Postgres, same approach as
  `live_grievances.result_json`), `num_shingles` (0 = abstained, per
  `minhash_signature()`'s contract), `script`/`window_index`/`block_key`,
  and the two identity keys. Upserted, batched, resumable.
- **`dedup_groups`** — one row per ticket: `duplicate_group_id` (union-find
  root, singleton = itself) and `group_size`. Recomputed and upserted whole
  on every grouping run rather than incrementally: a slice's signatures fit
  comfortably in memory, and a late-arriving resubmission can change a group
  id that already exists, so an incremental update would drift. Every
  ticket with a signature gets a group row (including singletons) so row
  counts can be reconciled against the slice definition, per #71 and to
  support #72's recall measurement against the ~34,700 officer-confirmed
  duplicates.

No third upsert helper — both tables upsert via `_dialect_upsert` imported
from `janasunani.pipeline.export`.

## Classification: deliberately not in `LAKE_TABLES`

ROADMAP §3.2 is explicit that redaction lowers exposure but does not
declassify what's derived from citizen prose — a MinHash signature or a
duplicate-group id can re-identify a filing by distinctive phrasing where a
phone number no longer can, and classifies dedup-index artifacts as
`dpic-infra`, same tier as the raw lake, never an `authorized-external`
route. I read that as **not** the same bar `grievance_redactions` clears
(which is safe for the lake because it's typed placeholders, no residual
signal). The Parquet lake also isn't access-controlled per table the way the
OLTP store is, and Phase 18's RBAC/audit scope (the control ROADMAP §3.2
actually names for these artifacts) hasn't landed. So `dedup_signatures`/
`dedup_groups` stay OLTP-only for now — commented in both `models.py` and
`materialize.py`'s `LAKE_TABLES` as a scope decision for #71, not a
permanent one.

## Salt handling

`DEDUP_SALT` (new `Settings` field, `.env.example` entry) or `--salt`.
`build_dedup_index()` raises `ValueError` — before opening any DB connection
— if neither is set, matching `identity_key()`'s own "fail loud on a
deployment-level problem, don't abstain" stance for a blank salt. Tested by
monkeypatching `create_async_engine` to raise if called, then asserting the
`ValueError` fires first.

## Timestamps

Naive UTC (`datetime.now(timezone.utc).replace(tzinfo=None)`) on both
`indexed_at` and `grouped_at`, same reasoning as `redact_grievance.py`
(#129: asyncpg rejects a tz-aware value into `TIMESTAMP WITHOUT TIME ZONE`
while SQLite silently accepts one). Both pinned by tests.

## ROADMAP

Flipped Phase 14's status from ⬜ to 🔄 with a note that the runner is
merged but no real slice has been run yet and spam_duplicate scoring is
still unbuilt; updated the backfill-order line in §5.2 to name the console
script.

## Verification (CI is down — this is the only evidence)

```
uv run ruff check .
  All checks passed!

uv run --extra serving --extra pipeline-core pytest
  593 passed, 14 skipped in 36.36s
```

This sandbox's `main` baseline (verified directly, not assumed) is **551
passed, 14 skipped** — not the 558/7 quoted in the issue, which looks like
it came from an environment with DVC-pulled master tables and a reachable
Postgres test DB. All 14 skips here are pre-existing, environment-gated
(`no reachable Postgres test DB`, `needs the DVC-pulled janasunani-mappings
master tables`, tesseract/GPU-model smoke) and identical in count on `main`
and on this branch — this diff adds exactly 42 passing tests and changes
nothing else.

Alembic: exactly one head.
```
$ OLTP_DB_URL=sqlite+aiosqlite:///.../scratch.db uv run alembic heads
e59fb4410dd6 (head)
```
Also verified upgrade -> downgrade -> upgrade round-trips cleanly on SQLite.

## Sanity-check for the maintainer

- The blocking window default (30 days) and duplicate threshold default
  (0.5, matching `tests/test_dedup.py`'s own end-to-end example) are my
  calls, not specified anywhere — worth a second look before a real slice
  runs.
- The `LAKE_TABLES` exclusion is my read of ROADMAP §3.2's classification
  language; flagging in case the intent was actually for these to ride
  along with `grievance_redactions`.
- Resubmission linking only spans the district-year slice processed in one
  invocation (a citizen refiling in a different year isn't linked) — noted
  in the module docstring, not fixed here.


ymohanty:
Reviewed the safety properties independently rather than taking the report on trust. All hold:

- `dedup_index.py` selects `GrievanceRedaction.grievance_redacted`; `Complaint.grievance` appears exactly once in the file, inside a docstring saying "never"
- Single alembic head `e59fb4410dd6`
- **593 passed, 14 skipped, ruff clean** on my machine, matching your numbers
- `block_key` is `district:script:window_index`, so cross-script separation is structural rather than a convention — Odia and Roman text cannot collide by construction
- `build_dedup_index()` refuses to run without a salt, which is the right failure mode

`test_identical_raw_text_with_different_redactions_does_not_group` is better than what the brief asked for. Identical raw text that redacts differently must not group, so the test fails immediately if anyone repoints the query at the raw column — it tests the property rather than the line.

## On excluding the tables from `LAKE_TABLES`

I checked the two consumers before commenting, and **your call holds**, with one consequence worth naming in review.

Both #72 and #73 can work entirely from OLTP: the lake is materialised *from* OLTP, so `complaints` and `action_history` are both available there, and #73 reads through the serving API anyway.

The friction lands on **#72**, which wants "prevalence by district, category, mode, year over the slice" plus recall against the ~34,700 officer-confirmed duplicates drawn from discard-reason templates in `action_history`. That is aggregate analytical work — exactly the workload the Parquet lake and duckdb exist for — and it would now run against Postgres instead. Workable at 55,544 rows, and it would not be at 1.37M.

So the trade is real but correctly weighted: **a slower analytical path is recoverable, an over-shared index is not.** ROADMAP §3.2 is explicit that redaction lowers exposure without declassifying what is derived, and distinctive phrasing re-identifies where a redacted phone number no longer does. Worth a line on #72 so whoever builds it knows to target OLTP and does not go looking for a Parquet file.

## The two defaults worth a second opinion

The 30-day window and the 0.5 threshold are your calls and they are reasonable, but they are the two numbers that move the headline duplicate rate — which is a DELIVERY.md Table 2 figure. Whatever they end up as, they belong in #72's artifact as stated parameters rather than as defaults nobody sees, since a reader cannot interpret a duplicate rate without them.

## Minor

You touched `docs/ROADMAP.md`. That file has had three separate edits today (#123, #124 merged, plus this), so worth a rebase check before merge even though the sections differ.

ymohanty:
Addressed the three fixes from review round 1 (leaving the two filed items alone):

**FIX 1 (P1, identity match != duplicate evidence).** `_group_duplicates` now runs identity-key candidate pairs through the same `jaccard_similarity` check as LSH text candidates before either is allowed to union — identity equality only widens candidate generation (across time windows/scripts), it no longer merges unconditionally. Added `test_same_mobile_different_issue_does_not_group` / `test_same_email_different_issue_does_not_group` (same identity, unrelated text -> not grouped) and kept positive coverage with two new fixture tickets (`T5R`, `T20R`: same identity + matching text, different window -> grouped) so the identity path is still proven to work, just gated. `identity_key_mobile`/`identity_key_email` already lived on `dedup_signatures` per-ticket, so "same citizen, different issue" stays queryable without any new column — added `test_same_citizen_different_issue_stays_queryable_via_identity_key` to pin that.

**FIX 2 (P2, star edges under-connect a bucket).** Both `_text_candidate_pairs` and `_identity_candidate_pairs` now emit every unordered pair via `itertools.combinations` instead of first-to-other edges (applied the same fix to identity groups too, since FIX 1 makes them per-pair-verified as well, so they have the same failure mode). Added `TestBucketVerifiesAllPairsNotJustStarEdges`, a direct unit test of `_text_candidate_pairs` with a hand-built 3-member bucket (real text collisions in a specific band aren't reliably constructible, so this drives the function with synthetic row objects rather than a full DB round trip).

**FIX 3 (P2, unvalidated threshold).** `build_dedup_index()` now rejects a non-finite or out-of-`[0, 1]` threshold before opening any DB connection, mirroring the existing salt check. Added `TestThresholdValidation` (negative, >1, NaN, and the 0/1 boundary values).

Left the two filed items untouched (salt-rotation/`--window-days` reindex lifecycle, and the lake-vs-OLTP design question).

## Verification

```
uv run ruff check .
  All checks passed!

uv run --extra serving --extra pipeline-core pytest
  601 passed, 14 skipped in 40.08s
```

That's the 551-passed/14-skipped `main` baseline (verified directly in this sandbox, see the earlier comment) plus 50 passing tests in `tests/test_dedup_index.py` (was 42; net +8 from this round: -3 old identity tests +6 new ones, +1 bucket test, +4 threshold tests). Skip count unchanged and still entirely pre-existing/environment-gated (no reachable Postgres, no DVC-pulled mapping tables, tesseract/GPU-model smoke).

Alembic: still one head (`e59fb4410dd6`), untouched by this round — no migration changes needed for these fixes.

Issue #108, opened 2026-08-06
Findings brief and deck for the external audience


ymohanty:
Parent: #51. Filed 6 August.

The artifact that leaves the room. The ED's point is that a live demonstration generates interest in the room and nothing afterwards; a written brief is what circulates.

Audience is **external government**: secretaries, departments, ORTPSA. Written for someone who will not ask a follow-up question.

### Done when
- [ ] Brief in `docs/FINDINGS.md`, rendered to DPIC-branded Word through the existing `scripts/md_to_docx.py` and `make docs`
- [ ] Deck through the `dpic-deck` skill
- [ ] Each finding: the number, its denominator, its caveat, what it implies. In that order
- [ ] Each labelled Insight or Capability explicitly

### Write it in this order, and the order is the point
State what an analyst with database access could have found, **first and plainly**. ROADMAP §5.3 is right that "yes, your DBA could have found that" is a fine answer given first and a bad one heard as a correction.

Then the two things no dashboard can produce: the duplicate increment, and a surge separated into filings, distinct problems and distinct citizens.

### Do not
- Do not lead with the pipeline, the PII scorecard or the deployment. They are appendix material now
- Do not report the closure figure as a failure rate
- Do not present the routing crosswalk as optimising for outcomes. It learns where cases were sent. See #106
- Do not report a Sarvam accuracy verdict. Comparative only, per #84

Depends on the findings pack. Gated on ED clearance for the closure finding (DELIVERY.md Table 4).

ymohanty:
## Verified — one gap found and fixed

- [x] **Brief in `docs/FINDINGS.md`** — present, 6,495 bytes.
- [x] **Rendered to DPIC-branded Word through `scripts/md_to_docx.py`** — verified, produces a 40,373-byte `.docx`.
- [x] **…and `make docs`** — this was **not** wired. `DOC_SOURCES` listed only `DELIVERY.md` and `ROADMAP.md`, so `make docs` never built the brief and the render was a manual step. Fixed in #166 (`e12b545`); `make -n docs` now emits the `docs/FINDINGS.docx` rule. Output stays gitignored, Markdown remains source of truth.
- [x] **Deck through the `dpic-deck` skill** — `docs/FINDINGS.pptx`, valid OOXML, 8 slides: DPIC title, Sprint 3 Findings, Insight 1-5, Next Evidence.
[...]

PR #165, opened 2026-08-07
feat(analytics): add five-script Sprint 3 findings pack

### LINKED ISSUE #107: Findings pack: five reproducible findings from the record
Parent: #51. **Filed 6 August**, on the ED's finding that the pipeline and deploy work does not interest an external audience.

The demonstration now opens with what the record says about how redressal works, and the software becomes the evidence behind it. This issue is that material.

### Why this is a one-day item
Most of it is already computed. The numbers below are in ROADMAP §5.2 and §5.3 as prose because somebody already ran the SQL ad hoc. What does not exist is a **reproducible, tested, presentable** version.

It reads `action_taken_remark` and structured `complaints` columns. **It never reads `complaints.grievance`.** So it needs no redaction pass, no dedup index, no gold set, no slice decision, no deploy and no GPU. That is the whole reason this fits.

### The five
| # | Finding | Kind | Depends on |
|---|---|---|---|
| 1 | Share of closures recording no action, both denominators | Insight | nothing |
| 2 | Cases created and closed within two days on a bare disposal | Insight | #75 |
| 3 | Why complaints get discarded, eight reason families | Insight | nothing |
| 4 | Duplicates the manual process already catches (~34,700) | Insight | nothing |
| 5 | Misrouting: 8,455 discarded as outside the office's remit | Insight | nothing |

Finding 2 is the new one and it is the most useful of the set, because it names a specific set of cases rather than indicting the system as a whole.

Findings 4 and 5 are baselines that set up capability claims made elsewhere: the dedup increment (#72) and the crosswalk (#33). Present them in that order or the capability reads as a restatement.

### Done when
- [ ] One script per finding under `scripts/analysis/`, over the existing lake read path (`janasunani/olap/lake.py`: `connect()` registers every Parquet as a DuckDB view, `query()` returns a DataFrame). Do not rebuild that plumbing
- [ ] Each writes a CSV or Parquet to `outputs/findings/` plus a Markdown fragment
- [ ] Each ships a test asserting its number against a **fixture**, never against the live lake
- [ ] Every headline number reconciled against a second, independently written query. The 61% already moved by half once depending on denominator
- [ ] Every output labelled **Insight** or **Capability** out loud, per the ROADMAP §5.3 value-add test

### Hard constraints
- ⚠️ **Aggregates only.** No script prints a row of citizen text, ever
- ⚠️ Exact-match lookup over high-frequency templates only. A string used 20,000 times is not personal data; the free-text tail is a privacy boundary as well as a scope one
- Every use of the 61% states its denominator in the same breath. Against all 1,209,138 resolved complaints it is 39.1%
- The closure figure is **descriptive and never a failure rate**. Correct and premature closure are identical in the record

Suitable for `executor-sonnet`.

ymohanty:
## Summary
- give all five #107 findings separate runnable scripts and aggregate CSV/Markdown artifact pairs
- preserve the governed closure SQL while reconciling it at runtime against an independently structured arg_max query
- use live-verified exact high-frequency templates for eight discard reasons, the officer-confirmed duplicate baseline, and the out-of-purview misrouting baseline
- retain the dated ROADMAP counts beside current counts and deltas so refreshes cannot rewrite the baseline silently
- narrow lake connections to explicit table allowlists; this stage opens only complaints and action_history and never reads complaints.grievance
- add a reproducible DVC findings stage and push its 10 aggregate outputs to the configured remote

## Live aggregate results (7 August snapshot)
- closure bare rung: 472,782 / 776,922 templated closures = 60.9%; 472,782 / 1,209,144 resolved = 39.1%
- created and closed within two days on a bare rung: 8,974; 1,020 on the minimum three-action trajectory
- officer-confirmed duplicate action rows: 37,299, +2,628 from the ROADMAP reference of 34,671
- out-of-purview action rows: 8,472, +17 from the ROADMAP reference of 8,455

## Safety
- accessed only the explicitly authorized data/interim/complaints.parquet and data/interim/action_history.parquet
- never selects complaints.grievance or ticket identifiers for publication
- never serializes action remarks; DVC outputs contain only governed labels and aggregate counts
- DVC reproduction used --single-item, so the upstream materialize stage did not run

## Verification
- DVC stage reproduced; dvc status findings: up to date; 10 files pushed
- uv lock --check
- uv run ruff check .
- full serving + pipeline-core suite: 844 passed, 18 skipped

## Scope
This completes the code, live reconciliation, and DVC artifact gates for #107. It intentionally does not close #107: assembly and clearance of the final external brief remain in #108.

Closes #107


ymohanty:

ruff — All checks passed
pytest — **844 passed, 18 skipped**

#166 is stacked on this branch (its base is now `feat/findings-pack`), so this must merge first.

Integration checked: all 11 distinct PRs merged together give ruff clean, **1002 passed, 14 skipped**, frontend build exit 0 (routes `/`, `/_not-found`, `/history`, `/supervisor`), eslint clean, 3 frontend tests passing.

Issue #92, opened 2026-08-05
NAME recall is 0.436: the NER misses 228 of 404 citizen names while over-predicting by 23%


ymohanty:
## The measurement

First scoring of Presidio against a structurally valid gold (#15, `ffdfd30`, gold md5 `efeca59d…`):

```
entity  gold  pred  overlap_hits  exact_hits  overlap_recall  exact_recall
NAME     404   497           176         109          0.4356        0.2698
```

**228 of 404 citizen names are missed.** Coverage overall is 0.4650 against the DSI English reference of 0.8056.

PHONE (0.966) and AADHAAR (0.857) are fine. EMAIL (0.371) is largely an artifact of this gold counting ~49 government addresses that #56 rules are not PII, and should reach ~0.775 once that lands. **NAME is the only entity whose gap nothing currently open explains.**

## This is not a threshold problem

The obvious lever is to lower the PERSON confidence threshold, since §5.1 says a recall-favouring setting with human review beats a balanced one. The numbers argue against it:

- **497 predicted against 404 gold** — the tagger already over-predicts names by 23%
- **only 176 of those 497 overlap a gold span** — 321 predictions hit nothing
- **exact_hits 109 vs overlap_hits 176** — even where it finds a name, it gets the boundary wrong more often than not

Simultaneous over-prediction and low recall means the model is looking in the wrong places, not being too cautious. Lowering the threshold adds noise to the 321 without recovering the 228.

The likely cause is `en_core_web_sm` on Indian text: tagging department names, place names and transliterated Odia as PERSON while missing actual Indian personal names. Worth confirming by sampling the 321 false positives and the 228 misses before choosing a fix.

## Why it matters more than the aggregate suggests

§5.1: the release-critical metric is the **false-negative rate**, because a missed name is leaked citizen PII, and F1 hides it. A missed PHONE is a leaked number; a missed NAME is a leaked identity, and it is the entity most likely to appear in a grievance.

The pipeline's whole safety claim rests on redaction working. At 0.436 it does not, for the entity that matters most.

## Options, in ascending cost

1. **Custom Indian-names recognizer** — gazetteer plus patterns, in the style of the existing `IN_MOBILE` / `IN_AADHAAR` / `IN_PAN` `PatternRecognizer`s at `pii_tagger.py:115-153`. No new heavy dependency, no `transformers` version conflict, agent-suitable. Most likely to fit before 14 August.
[...]

Issue #53, opened 2026-07-27
Phase 17 · Sarvam benchmark, provider registry, egress control


ymohanty:
Component **(e)** of the five-component demo re-scope (2026-07-27). Ref: [ROADMAP.md §5.5](../blob/main/docs/ROADMAP.md) and [§3.1](../blob/main/docs/ROADMAP.md).

Sarvam is the most promising single quality lever available: our weakest stage and their strongest capability are the same thing — **Odia document understanding**.

## ⚠️ This phase retires the no-egress invariant
The Government of Odisha has authorized Sarvam for this data **including PII**. "Citizen text never leaves the box" becomes **"citizen text leaves the box only through a declared, audited, revocable channel."**

This **tightens** enforcement in practice: today no-egress is policy, not network — the boxes allow general outbound and BART downloads from a public hub at startup. An allowlist permitting Sarvam and nothing else is stricter than the status quo.

### Sarvam capabilities (verified 2026-07-27 — re-check, this moves fast)
- **Sarvam Vision** — 3B state-space VLM for document digitisation, 22 Indian languages + English. Replacement candidate for pytesseract-`ori` and DeepSeek-OCR.
- **Sarvam-30B** (MoE) and **Sarvam-105B** (~9B active, 128k ctx) — **open-weight, Apache 2.0** since Feb 2026.
- **Sarvam Translate** (22 scheduled languages) and **Mayura** (colloquial / code-mixed). Romanized Odia is exactly a code-mixed problem.
- **Saaras v3** ASR with a transliterate output mode.

### The open weights are the architectural point
Sarvam is a **provider with two backends**, not an external dependency:
- `sarvam-hosted` — the API under the Odisha authorization. Cheap, no GPU, fastest path to a benchmark.
- `sarvam-selfhosted` — Apache-2.0 weights on the GPU box. No egress at all.

Both behind **one registry entry**. That is what makes the egress decision reversible.
- [ ] ⚠️ **Measure before assuming the exit ramp is usable** — Sarvam-30B at 4-bit is roughly at the limit of the current 24 GB L4; 105B is beyond it and would need a larger instance type.

### Registry generalization
A model reference becomes `{name, alias, provider, artifact_or_endpoint, version, egress_class}`.
[...]

PR #169, opened 2026-08-07
feat(evaluation): Sarvam Vision vs pytesseract scorecard — paired, clustered, split per #84 #127

### LINKED ISSUE #84: Sarvam Vision vs pytesseract scorecard, printed and handwritten separately
Parent: #53. The deliverable for component (e). DELIVERY.md: "the scorecard is the
deliverable and adoption follows the result".

Sarvam Vision against our pytesseract extraction, on the hand-transcribed sample.

### Done when
- [ ] Both systems scored against the same transcription
- [ ] **Handwritten and printed reported separately**
- [ ] Odia and English reported separately

⚠️ Sarvam documents their vision model as strong on printed text, tables and layout,
with **no mention of handwriting**. Much of our corpus is handwritten. A result that
holds only for printed forms is a much smaller result than it appears, and reporting
one blended number would hide exactly that.

**Blocked on the transcription owner.** If none is named, this ships in its stated
fallback form: how the two systems differ from each other, with no verdict on which
is right, and the accuracy row dropped from DELIVERY.md Table 2.

Sarvam-30B on categorisation and summarisation, and the provider switch, remain
stretch per DELIVERY.md and are not part of this issue.

### LINKED ISSUE #127: Sarvam comparison: paired small-sample design, estimator and power
The Sarvam comparison is a small paired sample against our current open-source pipeline, both reported with standard errors. Not a corpus run (#83: Doc AI is a measurement instrument, ~₹145,000 and ten days of calling over the full corpus).

This issue fixes the design **before** any page is submitted, because most of the ways this goes wrong are choices made after seeing results.

## ⚠️ The error that would invalidate the headline

**Do not compare Sarvam's sample number against MuRIL's 0.7104.**

That figure comes from a 65,999-grievance test set of **typed subject lines**. The Sarvam sample is **scanned pages**. Comparing them measures the difference between typed text and scans at least as much as the difference between two models, and the confound is not separable after the fact.

**Re-run our pipeline on the same pages.** Every number in the comparison must come from the same sample. MuRIL's 0.7104 is context in the write-up, never a term in the comparison.

## The design

**Paired.** Both systems process the same pages. This is the decision that makes a small sample viable: page difficulty is the dominant variance component, and pairing removes it entirely rather than averaging over it.

**Report the confidence interval on the difference, not two intervals.** Two overlapping CIs do not mean no difference — for paired data the interval on the difference is much tighter than either marginal interval, and reading overlap as "no effect" is the standard way a real difference gets missed. The difference and its CI is the result; the two marginal rates are description.

**Cluster standard errors by document, not by page.** Pages from one scanned complaint share handwriting, scan quality and subject matter, so they are not independent draws. Treating 200 pages from 40 documents as n=200 understates the standard error by roughly the square root of the cluster size. Cluster on ticket.

**Pre-register the primary outcome.** One primary comparison, everything else exploratory and labelled as such. Per-class category results across a dozen classes will produce a significant result by chance; that is fine to explore, not to headline.

## Sample size

Paired (McNemar), two-sided 95%, 80% power. Discordance is the share of pages where the two systems disagree.

| Gap to detect | 20% discordance | 30% discordance |
|---|---|---|
| 5 points | 628 pages | 941 pages |
| 10 points | 157 pages | 236 pages |
| 15 points | 70 pages | 105 pages |
| 20 points | 40 pages | 59 pages |

At ₹1.00/page for Extract, **even the 5-point design costs about ₹950.** Cost is not the constraint; assembling and labelling the sample is.

For contrast, an *unpaired* comparison at p≈0.71 gives a CI half-width of **±12.6 points at n=50** and ±6.3 at n=200. A 50-page sample cannot distinguish 0.71 from 0.84 unpaired. **Pairing is what makes a small sample worth running at all.**

**Recommendation: 200-250 pages, stratified by category**, powered for a ~10-point gap. Stratify so Social-Welfare — MuRIL's weakest class at 0.51 F1, and the one where a general Indic model is most likely to win — is thick enough to say anything about. Proportional sampling would leave it too thin.

## What has a referee and what does not

| Comparison | Reference | Reportable as |
|---|---|---|
| **Category** | recorded category on the ticket | **A real accuracy comparison with CIs.** The headline |
| **PII detection** | the n50 gold, 529 spans over 89 records | A comparison, with caveats below |
| **Transcription** | none — #65 deferred | **Divergence only.** No verdict on which is right |

Transcription has no referee, so there is no "our result versus theirs" to compute. A disagreement rate with a CI is a descriptive statistic, and presenting it beside two comparisons that do have references will invite it to be read as a third. Separate it visually and say what it is.

PII is worth attempting but is the weakest of the three: n=89 records once clustered, and doc-ai returns field *values* while the gold holds *spans*, so a matcher is needed and its leniency moves the result (#92). Report it as a secondary finding with the matching rule stated, or not at all.

## Done when

- [ ] Sample drawn and frozen, stratified by category, ticket ids recorded
- [ ] Our pipeline re-run on **those pages**, not quoted from a previous evaluation
- [ ] Paired estimator with ticket-clustered standard errors
- [ ] Primary outcome named in writing before submission
- [ ] Markdown-to-text normaliser for the transcription arm frozen before any output is read (#84)
- [ ] Result reported as difference plus CI, with marginal rates as description

## Related

- #53 — Phase 17
- #83 — the adapter, per-page pricing, and why this is not a corpus run
- #84 — the transcription arm, divergence-only
- #126 — the category arm, which this supersedes on design
- #92 — the PII arm and its value-versus-span matching problem
- #81 — the Phase 16 A/B power calculation; different experiment, same estimator discipline

ymohanty:
Closes #84. Addresses #127.

## Summary

Paired Sarvam Vision vs pytesseract scorecard with ticket-clustered CIs per #127, stratified handwritten/printed and Odia/English splits per #84. Primary outcome is pre-registered as category accuracy difference (the only arm with a referee — recorded category on the ticket); transcription is reported as divergence-only until a hand-transcribed ground truth exists, at which point the same harness reports accuracy. Verified with recorded (synthetic) responses only — no live Sarvam calls.

## Design decisions (frozen before submission, #127)

**Paired.** Both systems process the same pages. Page difficulty is the dominant variance component; pairing removes it rather than averaging over it. This is what makes a 200–250 page sample powered for ~10 points viable (McNemar; see `required_pages_mcnemar`). An unpaired comparison at p≈0.71 cannot distinguish 0.71 from 0.84 at n=50.

**Primary outcome pre-registered.** One headline: *difference in category accuracy (Sarvam Extract vs pipeline categorizer) on the paired sample, ticket-clustered 95% CI*. All per-class category results and both transcription results are exploratory and labelled as such — a dozen per-class tests will produce a significant result by chance.

**Report the CI on the difference, not two intervals.** Two overlapping marginal CIs do not imply no difference for paired data; the interval on the difference is tighter and is the result. Marginal rates are description.

**Cluster SE by ticket, not page.** Pages from one scanned complaint share handwriting, scan quality and subject matter and are not independent. The harness uses a sandwich estimator: `Var = sum_c (sum_{i in c} (d_i - mean))^2 / n^2 * C/(C-1)`. Treating 200 pages from 40 documents as n=200 understates SE by ~sqrt(cluster size).

**Handwritten vs printed reported separately (#84).** Sarvam state Vision is trained on handwritten text across 22 languages but publish no handwriting-specific number; much of the corpus is handwritten. A blended number would hide the exact question the corpus turns on. Split uses `pages.handwritten` (`yes`→handwritten, `no`→printed, `partial`/`NULL`→separate bucket) and is reported for both divergence and, when ground truth exists, accuracy.

**Odia vs English reported separately (#84).** Same rationale — language is the other axis the benchmark exists to measure.

**Markdown-to-text normaliser frozen before any output is read (#84, #127).** Digitise returns Markdown, pytesseract returns plain text; comparing without normalisation measures markdown headedness. Normaliser `NORMALIZER_VERSION=1.0` strips headings/bold/italic/code/links/images/tables, NFC-normalises, collapses whitespace, and is versioned so its aggressiveness cannot be tuned after seeing results.

**Same-sample pipeline re-run.** The comparison re-runs the local pipeline on the same pages; it never compares a Sarvam sample number against MuRIL's 0.7104 (typed subject lines on 65,999 grievances), which would confound model with typed-vs-scan (#127).

## Transcription sample — flag

**No hand-transcribed sample is available.** Per DELIVERY.md / #53 / #84, no owner was named for the ~50-page (printed+handwritten, Odia+English) transcription set that the OCR accuracy comparison requires. The previous 77.9% was a plausibility pass rate, not accuracy, so there is no ground truth to reuse, and an agent cannot produce it because it *is* the ground truth an agent would be measured against.

**This PR ships the stated fallback (DELIVERY.md: "the scorecard is the deliverable and adoption follows the result"):** the transcription arm reports **divergence only** — how the two systems differ after the same normalisation, with ticket-clustered CI — with no verdict on which is right, and the accuracy row is dropped from DELIVERY.md Table 2. The harness flags this explicitly (`transcription_available=False`, `needs_transcription_sample=True`, and a note in `ScorecardReport.notes`). When a sample is commissioned, the same `build_scorecard` call will report accuracy; no code change is needed.

Stratification note per #127: the recommended 200–250 page sample should be stratified by category so Social Welfare (MuRIL weakest at 0.51 F1) is thick enough to say anything about; proportional sampling would leave it too thin. Sample, once drawn, must be frozen and its ticket ids recorded.

## What has a referee

| Comparison | Reference | Reportable as |
|---|---|---|
| Category | recorded category on ticket | Real accuracy comparison with CI — headline |
| Transcription | none until sample commissioned | Divergence only (descriptive) |

## Verification

- `ruff check janasunani/evaluation/sarvam_scorecard.py tests/test_sarvam_scorecard.py` — All checks passed
- `pytest tests/test_sarvam_scorecard.py -v` — 8 passed (recorded/synthetic fixtures only; no live egress)
- No `data/` access, no network calls, `NORMALIZER_VERSION` recorded in artifact

## Follow-up

Commission the hand-transcribed sample (owner to be named) to make the transcription accuracy comparison reportable. The PII arm (n=89 once clustered, value-vs-span matching per #92) is not in scope for this PR.

Closes #127

### DIFF
--- a/janasunani/evaluation/__init__.py
+++ b/janasunani/evaluation/__init__.py
@@ -1,0 +1,2 @@
+"""Evaluation harnesses that score pipeline stages against ground truth."""
+
--- a/janasunani/evaluation/sarvam_scorecard.py
+++ b/janasunani/evaluation/sarvam_scorecard.py
@@ -41,6 +41,7 @@ import re
 import unicodedata
 from collections import defaultdict
 from dataclasses import asdict, dataclass
+from statistics import NormalDist
 from typing import Any
 
 NORMALIZER_VERSION = "1.0"
@@ -274,10 +275,17 @@ def required_pages_mcnemar(gap: float, discordance: float, alpha: float = 0.05, 
     """
     if not (0 < gap < 1 and 0 < gap < discordance <= 1):
         raise ValueError("need 0 < gap < discordance <= 1")
-    # z critical values
-    z_alpha = 1.96 if alpha == 0.05 else 1.96
-    # 80% power -> z=0.84
-    z_beta = 0.84 if power == 0.8 else 0.84
+    if not 0 < alpha < 1:
+        raise ValueError("alpha must be in (0, 1)")
+    if not 0 < power < 1:
+        raise ValueError("power must be in (0, 1)")
+    # Actual quantiles, not the 0.05/0.80 constants. Both ternaries here
+    # returned the same value on either branch, so asking for 90% power or a
+    # 1% level silently returned the 80%/5% sample size — a deliberately
+    # under-powered study reporting itself as powered, which is the failure
+    # #127 exists to prevent, and every page submitted costs money.
+    z_alpha = NormalDist().inv_cdf(1 - alpha / 2)
+    z_beta = NormalDist().inv_cdf(power)
     # McNemar: n = (z_alpha*sqrt(p_d) + z_beta*sqrt(p_d - gap^2))^2 / gap^2
     p_d = discordance
     numerator = (z_alpha * math.sqrt(p_d) + z_beta * math.sqrt(max(p_d - gap * gap, 1e-9))) ** 2
--- a/janasunani/evaluation/sarvam_scorecard.py
+++ b/janasunani/evaluation/sarvam_scorecard.py
@@ -1,0 +1,482 @@
+"""Sarvam Vision vs pytesseract scorecard — paired design per #127, fallback per #84.
+
+Design decisions (frozen before submission, see #127):
+
+* Paired: both systems process the same pages. Page difficulty is the dominant
+  variance component and pairing removes it rather than averaging over it.
+* Primary outcome (pre-registered): difference in **category accuracy** on the
+  paired sample, because it has a referee (recorded category on the ticket)
+  and the transcription arm has none. Everything else is exploratory and
+  labelled as such. Per-class results are exploratory; the headline is the
+  overall difference.
+* Report the confidence interval on the **difference**, not two marginal
+  intervals. Two overlapping CIs do not imply no difference for paired data.
+* Cluster standard errors by **ticket (document)**, not by page. Pages from
+  one complaint share handwriting, scan quality and subject matter.
+  Treating 200 pages from 40 documents as n=200 understates SE by ~sqrt(5).
+* Handwritten vs printed reported **separately** (issue #84). Sarvam documents
+  Vision as strong on printed text/tables/layout; handwriting is claimed but
+  has no published handwriting-specific number. A blended number would hide the
+  exact question the corpus turns on. Split on the ``handwritten`` column
+  (``yes`` = handwritten, ``no`` = printed, ``partial``/NULL = separate bucket).
+* Language split: Odia and English reported separately (issue #84). The corpus
+  is Odia-heavy; a blended figure hides the Odia question entirely.
+* Transcription arm: if no hand-transcribed ground truth is available (the
+  current state per DELIVERY.md — no owner was named), report **divergence only**
+  — how the two systems differ from each other — with no verdict on which is
+  right, and drop the accuracy row from DELIVERY.md Table 2. This module flags
+  that condition explicitly (``transcription_available`` / ``needs_transcription_sample``).
+* Markdown-to-text normaliser is frozen **before any output is read** (#84,
+  #127). Both engines are compared after the same normalisation so markdown
+  headedness does not move the headline. Version is recorded in the report.
+
+Verification: recorded responses only. Tests supply synthetic page records and
+recorded Sarvam markdown fixtures; no live network call is made.
+"""
+
+from __future__ import annotations
+
+import math
+import re
+import unicodedata
+from collections import defaultdict
+from dataclasses import asdict, dataclass
+from typing import Any
+
+NORMALIZER_VERSION = "1.0"
+
+# ---------------------------------------------------------------------------
+# Normaliser — frozen before output is read
+# ---------------------------------------------------------------------------
+
+_MD_TABLE_RE = re.compile(r"^\s*\|.*\|\s*$", re.MULTILINE)
+_MD_HEADING_RE = re.compile(r"^\s{0,3}#{1,6}\s+")
+_MD_BOLD_RE = re.compile(r"\*\*([^\*]+)\*\*")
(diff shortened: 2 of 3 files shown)


ymohanty:

`uv run --extra serving --extra pipeline-core ruff check .` — All checks passed
`uv run --extra serving --extra pipeline-core pytest` — **835 passed, 18 skipped**

Merges cleanly with every other open Sprint 3 PR.

Integration checked: all 11 distinct PRs merged together give ruff clean, **1002 passed, 14 skipped**, frontend build exit 0, eslint clean.

ymohanty:
## Pushed a correctness fix — `e6a3742`

`required_pages_mcnemar` accepted `alpha` and `power` but ignored them: both z-quantiles came from ternaries returning the same value on either branch. Correct at the defaults, wrong everywhere else — `power=0.9, alpha=0.01` returned **194** pages where **368** are required, a study under-powered by 47% reporting itself as powered.

Now uses real normal quantiles and rejects out-of-range `alpha`/`power`. The previous test only called the function at its defaults; the new one asserts that tightening alpha or raising power strictly increases the requirement, and pins 259 and 368.

Also verified for #127: paired estimator on the categorisation arm, ticket-clustered SEs (`n_clusters=2` for 2 tickets × 2 pages, not n=4), the overlapping-CI warning, and no reference to MuRIL's 0.7104 anywhere in the estimator.

Re-verified: ruff clean, **933 passed, 14 skipped**. Body now also closes #127.

PR #170, opened 2026-08-07
docs: adopt the packaged DPIC agent standards

ymohanty:
## Why

The repo had no contribution guide, and `AGENTS.md` carried the gate commands
inline with nothing stating the commit, PR, subagent or review-finding
conventions. The org already packages those conventions in `dpic-org`, but the
`dpic` pin here resolved to `6c16b3e`, which predates that work, so
`dpic-sync-standards` did not exist in the venv and they could not be consumed
at all.

`sams` mirrors the same conventions inline by hand because it is pinned to
Python 3.10 and `dpic` requires >=3.13. This repo is on 3.13, so it takes the
real sync and never has to merge upstream wording manually.

## What

Two commits.

**`build(deps)`** — bump `dpic` to `dpic-org` main (`b7790bb`) and sync. Lands
`.dpic/standards/agent-conventions.md`; `.opencode/skills/dpic-deck` was already
identical, so no diff there. The synced tree is consumed verbatim, never
hand-edited: `dpic-sync-standards --check` reports drift, and changes are
proposed upstream.

**`docs`** — one authoritative home per topic:

- `CONTRIBUTING.md` (new): pre-PR checks, the Presidio-gated PII suite, the gold
  gates, DVC dependency rules, PR size, and the review protocol.
- `.dpic/standards/agent-conventions.md` (synced): everything org-wide.
- `AGENTS.md`: what neither covers, plus pointers to both. Dropped the
  duplicated gate commands.

The review protocol is spelled out because agents open PRs here: self-review the
diff, request a Codex review with `@codex review` on anything touching
pipeline/serving/deploy/PII/data, verify each finding before acting on it per
the synced standard, and file valid-but-not-blocking findings as tech-debt
issues rather than fixing them in the branch. No finding may be open at merge:
fixed, filed, or rejected with evidence in the reply. `@codex review` is the
trigger the connector actually listens for, confirmed from the comment history
on #29 rather than assumed.

Wiring:

- `CLAUDE.md` becomes a symlink to `AGENTS.md`, replacing a 14-byte
  `Read AGENTS.MD` stub that had drifted from the file it pointed at.
- `.claude/skills -> ../.opencode/skills`, so the synced `dpic-deck` skill loads
  from the repo. It was only reaching Claude Code through a stale copy in
  `~/.claude/skills`, which no clone would have.
- `make standards` re-syncs; `make setup` calls it.
- `.gitignore`: `.worktrees/`, which the conventions prescribe and require an
  ignore rule for, plus `.muse/worktrees/` which was showing up untracked.

Both symlinks are stored as mode `120000`, so they survive a clone.

## Verification

```
uv lock --check                     -> ok
uv run ruff check .                 -> All checks passed!
uv run dpic-sync-standards --check  -> DPIC skills and standards are in sync.
make standards                      -> re-syncs both trees, no diff
diff .dpic/standards/agent-conventions.md \
     ../dpic-org/dpic/standards/agent-conventions.md   -> identical
python -c "from dpic.documents import build_brief; from dpic.branding import colors"
                                    -> ok
```

`scripts/md_to_docx.py` holds the only `dpic` import in-tree, so the version
bump has no other reachable surface. No source files change; the gold gates and
the PII suite are not affected by this branch.

### DIFF
--- a/.claude/skills
+++ b/.claude/skills
@@ -1,0 +1,1 @@
+../.opencode/skills
--- a/.dpic/standards/agent-conventions.md
+++ b/.dpic/standards/agent-conventions.md
@@ -1,0 +1,123 @@
+# DPIC Agent Conventions
+
+Canonical source: `dpic/standards/agent-conventions.md` in `dpic-org`.
+
+**This file is synced — do not edit it in a project repo.** Local edits are
+reported by `dpic-sync-standards --check` and overwritten on the next sync.
+Propose changes against `dpic-org`.
+
+Project-specific instructions belong in the repo's own `AGENTS.md`, which
+should reference this file rather than restate it.
+
+## Attribution
+
+**Never mark commits as co-authored by an AI agent, or pull requests as
+authored by one.** No `Co-Authored-By: Claude ...` trailer, no
+`🤖 Generated with Claude Code` footer. This applies to every commit and every
+PR body without exception.
+
+## Commits
+
+- **Conventional Commits** subjects, optionally scoped: `fix:`, `feat:`,
+  `refactor:`, `build:`, `docs:`, `chore:`, `test:` — e.g. `docs(isi): ...`.
+- One commit per logical unit. A unit may span files; a file may hold several
+  units. Do not lump unrelated changes together.
+- Issue trailers: `Closes #NN` when the commit fully resolves the issue,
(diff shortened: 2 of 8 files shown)


ymohanty:
@codex review

PR #173, opened 2026-08-07
feat(analytics): duplicate recall baseline and prevalence (#72)

ymohanty:
## Summary
Adds the officer-confirmed duplicate baseline for #72 (Sprint 3 component b). The ~34,700 duplicates officers already caught are queryable today with a `CASE WHEN` — they are the baseline, not the deliverable. The capability claim is the increment MinHash finds that carries no such remark, kept separate by construction.

## What Changed
- `janasunani/analytics/sql/duplicate_recall.sql`: `duplicate_recall` mart — normalized exact-match over `action_history.action_taken_remark` (lower + whitespace collapse + trailing dot) for two families (`taken_up` 19,904 + `duplicate_copy` 14,767 = 34,671 ROADMAP ref), plus prevalence by district/category/mode/year and duplicate-adjusted workload. Reads `complaints` + `action_history` only; never reads `complaints.grievance` (asserted like `test_closure_finding` strips comments + string literals).
- `janasunani/analytics/findings/duplicate_recall.py`: `compute`/`render_markdown`/`main` mirroring `closure.py` — writes 6 CSVs + `duplicate_recall.md` to `outputs/findings/`, both denominators and ROADMAP delta, prevalence tables, workload.
- `tests/test_duplicate_recall.py`: 3 tests on fixture lake (no live data) — counts after normalization, grievance guard, compute+render headline.

## Decisions
- **Exact normalized match, not LIKE without normalization.** Officer remarks are picked from a dropdown, so the templated families are a structured signal wearing a text costume — same rationale as `closure.sql` ladder. Normalization is lower + single-space + strip trailing dots.
- **Four variants for taken_up:** `case already taken up`, `case taken up earlier`, `grievance already taken up`, `complaint already taken up` — all observed in dump; collapsing to one would undercount ROADMAP by ~200.
- **Baseline and increment stay separate.** `duplicate_baseline_summary` + prevalence views are baseline only; increment (MinHash duplicates without such remark) is a separate future view, not a column folded into share_pct. Reporting them together would present the baseline as the capability.

## Needs input
- **MinHash increment** — requires dedup index build (`janasunani.pipeline.dedup` + `dedup_index.py`) over the lake to report recall against the baseline and the true increment as the actual claim. This PR intentionally does not claim it.
- **Discard-rate variance by office** — #72 warns the confirmed set inherits the old process's bias; prevalence by district/category is reported but should not be read as ground truth without that check.

## Validation
- [x] `ruff check janasunani/analytics/findings/duplicate_recall.py tests/test_duplicate_recall.py` — All checks passed (fixed 1 unused import)
- [x] `pytest tests/test_duplicate_recall.py -v` — 3 passed (fixture lake, no data/ access, no network)
- [x] No `complaints.grievance` read (static + no column leakage)
- [ ] `uv run ruff check .` / full suite via `uv run` blocked in this env by `failed to open /Users/ymohanty/.cache/uv/sdists-v9/.git: Operation not permitted` — verified via direct `ruff`/`pytest` binaries instead

## Review Expectations
- Keep small: 3 files, ~300 lines.
- Yashaswi mandatory.
- Data And Delivery: does not modify data; SQL is the hand-over artifact.

Refs #72. Stacked on main 07be1e8.

## Data And Delivery Notes
- [x] This PR does not modify data.
- [ ] This PR ingests or versions data through DVC.
- [x] This PR updates generated exhibits.


ymohanty:

`uv run --extra serving --extra pipeline-core ruff check .` — All checks passed
`uv run --extra serving --extra pipeline-core pytest` — **830 passed, 18 skipped**

Merges cleanly with every other open Sprint 3 PR.

Integration checked: all 11 distinct PRs merged together give ruff clean, **1002 passed, 14 skipped**, frontend build exit 0, eslint clean.

ymohanty:
## Pushed a correctness fix — `8e38c08`

Verifying against the record turned up a predicate bug in the baseline this PR publishes.

`closing_remark IN ('case already taken up', ...)` matched **8** rows. The stored templates carry suffixes — `case already taken up for examination` (11,736) and `case taken up earlier hence closed` (9,289) — so the shipped finding reported **18,432** officer-confirmed duplicates (1.5%) and framed the gap against the ROADMAP's 34,671 as a data finding ("delta −16,239"). It was not.

Corrected by prefix-anchoring: **39,937** (21,513 taken up + 18,424 duplicate copy), **3.3%** of the 1,209,144 resolved complaints with a closing remark. Delta over ROADMAP is now **+5,266**, upward, and the rendered finding explains why rather than smoothing it — the reference query dropped ticket-referencing variants like `duplicate copy of DEPT2024669445`.

The prefix anchor also keeps the deferral template (`... kept in priority category and shall be taken up after due government approval`, 1,357 rows) out of the count, which a bare `LIKE '%taken up%'` would have swept in.

**The test would not have caught this.** The fixture held only `Case Already Taken Up.`, which matches both predicates. It now carries the real stored forms plus the deferral and fails `4 != 6` against the old SQL.

Re-verified: ruff clean, **927 passed, 14 skipped**.

Full detail on #72.

PR #236, opened 2026-08-10
fix: stop holding a write transaction open across a batch of Presidio calls

ymohanty:
## Summary

`_redact_slice` (`janasunani/pipeline/redact_grievance.py`) ran up to 500
sequential Presidio calls inside `async with engine.begin()`, holding a
write transaction open on the OLTP box for the whole batch. That box also
hosts production Postgres (1,371,288 real citizen complaints), so a slow or
interrupted batch held a write transaction open against the same server.

- Build the `rows` list (including the Presidio calls) before opening the
  transaction, so the transaction only performs the upsert.
- Left `BATCH_SIZE` and the per-row (non-batched, non-parallel) Presidio
  call alone -- parallelising it is separately blocked on the `pii` /
  `pipeline-core` numpy ABI conflict.
- Did not add an explicit yield to the event loop between batches: this
  script's own event loop has no other task running concurrently
  (`asyncio.run()` over a single coroutine), so a yield here would be a
  no-op given the current structure. Flagging in case that assumption
  should be revisited alongside actually parallelising Presidio.
- Redaction behaviour is unchanged -- this is a scheduling change only.

## Test plan

- [x] New test asserts redaction never runs while an `engine.begin()` block
      is open (tracks transaction depth against the real async engine, calls
      `_redact_slice` directly so it exercises the actual code path).
- [x] New test simulates a crash between building `rows` and the upsert
      (mid-batch, after Presidio finished, before the write transaction
      opens) and asserts the next run reprocesses the whole batch with no
      partial writes -- proving the NOT EXISTS resume predicate still holds.
- [x] New test asserts the persisted `grievance_redacted` output is
      byte-identical to calling the redactor directly on the same input.
- [x] `uv run ruff check .` -- clean.
- [x] `JANASUNANI_MODELS_DIR=/nonexistent uv run --extra serving --extra
      pipeline-core pytest` -- 1272 passed, 15 skipped, 0 failed.
- [x] `uv run --extra pii pytest tests/test_pii_extra_contract.py
      tests/test_pii_redaction.py tests/test_redact_grievance.py
      tests/test_rederive_pii_draft.py tests/test_bootstrap_pii_gold.py` --
      129 passed (including the real-Presidio test on this file, which
      skips in the serving/pipeline-core env).

### DIFF
--- a/janasunani/pipeline/redact_grievance.py
+++ b/janasunani/pipeline/redact_grievance.py
@@ -223,24 +223,35 @@ async def _redact_slice(
             batch = await _load_pending_batch(
                 conn, district, year, size, version=version if refresh_stale else None
             )
-            if not batch:
-                break
+        if not batch:
+            break
 
-            # Naive UTC. Every timestamp column in this schema is TIMESTAMP
-            # WITHOUT TIME ZONE, and asyncpg refuses to bind a tz-aware value
-            # into one while SQLite silently accepts it -- so an aware value
-            # here passes every local test and fails on the first batch against
-            # the deployed Postgres. Same normalisation as db/crud.py.
-            now = datetime.now(timezone.utc).replace(tzinfo=None)
-            rows = [
-                {
-                    "ticket_no": ticket_no,
-                    "grievance_redacted": redact(text),
-                    "redacted_at": now,
-                    "analyzer_version": version,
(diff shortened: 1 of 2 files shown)


ymohanty:
@codex review

PR #278, opened 2026-08-23
fix(ci): recognize structured Codex review commits

ymohanty:
## Summary

- recognize the current Codex review format through GitHub's structured `commit_id`
- require a full 40- or 64-character object ID from that authoritative field
- retain the legacy `Reviewed commit:` body parser for older reviews

## Why

Codex now posts findings without the legacy `Reviewed commit:` sentence. GitHub still records the exact reviewed commit in the review payload, but the gate ignored that structured field and falsely reported that reviewed PR heads had no Codex activity.

This remains fail-closed: only reviews authored by the recognized Codex identities count, an abbreviated structured ID is rejected, and unresolved Codex threads continue to fail the gate. The clean thumbs-up path and documented override are unchanged.

## Verification

- `python -m pytest tests/test_check_codex_review.py -q` — 43 passed
- `ruff check . --exclude data` — passed
- full offline suite — 1,466 passed, 22 skipped

The full suite used task-scoped caches with Hugging Face and Transformers forced offline so model initialization could not wait on network access.

### DIFF
--- a/scripts/check_codex_review.py
+++ b/scripts/check_codex_review.py
@@ -3,8 +3,10 @@
 Codex in GitHub reports neither a check run nor a commit status, so there is
 nothing to mark required in branch protection. It leaves two signals instead:
 
-- If it has findings, it posts a review whose body carries
-  ``**Reviewed commit:** `<sha>` `` and one inline thread per finding.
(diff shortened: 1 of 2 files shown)


ymohanty:
@codex review

PR #309, opened 2026-08-25
docs(presentations): add the analytics slide, draw the real-document slice

ymohanty:
## Summary

- Draws the real-document sample this branch's harness needs: nests the existing 92-ticket Sarvam sample inside a larger 224-ticket, 451-page draw over Sambalpur/2024 (30 categories), staged to scratch, outside any checkout. Nesting verified by ticket id and by file.
- Re-runs the three intelligence findings (`janasunani-publish-workload`, `-intelligence`, `-themes`) against real data reachable from this environment, and reports honestly where they do not reach the governed 55,544-filing slice (see notes on slide 8).
- Fixes a real crash in `janasunani.analytics.findings.themes.render_markdown`: it assumed the full `{theme_id, filings, ...}` schema, but `compute_themes`' "too few rows in this category" and "insufficient vocabulary" early exits return a `{category, note}` frame instead, so the CLI raised `ColumnNotFoundError` on a small local slice instead of reporting a graceful NULL. Tests added for both early-exit shapes plus the full render path (`tests/test_themes.py`).
- Adds slide 8, "The dashboards count filings, not problems," between the routing-outcome slide and "What we cannot measure," per docs/DELIVERY.md's intelligence-layer section (duplicate-adjusted workload, spike decomposition, local issue themes).
- Adds a provenance note to slide 4 ("Six stages, timed and scored"): document latency is one draw from Sambalpur 2024, the dedup slice. The latency numbers themselves are untouched — the real-document benchmark run was deliberately deferred to avoid CPU contention with a concurrent Sarvam evaluation on the main checkout.

## Branch note

This branch was cut from `feat/benchmark-real-documents`, which forked from the technical-briefing deck's first commit and never picked up the eleven commits that followed on `docs/timing-quality-briefing` (the routing-outcome slide, the six-stage timing table, the Sarvam head-to-head). The analytics slide this PR adds slots in after "The estimate did not survive the second year" and before "What we cannot measure," neither of which exist without that later work, so the first commit here (`merge(deck): ...`) brings the current deck structure over via path checkout (not a merge commit) — `scripts/benchmark_pipeline.py` and its test are untouched.

## What did not get done, and why

- **The latency benchmark itself was not run.** Explicitly deferred: a Sarvam evaluation was running local pytesseract on the main checkout, and any latency measurement taken during that run would be CPU-contention noise. The 224-document real sample is staged and ready (`sample_manifest.json` records seed, caps and per-category counts) for whenever the machine is free.
- **The governed 55,544-filing workload/spike decomposition was not reproduced locally.** The only OLTP dedup index reachable from this worktree (outside the deployment box) is a 20-ticket Sambalpur/2024 smoke-test population — real, digest-verified, but not the governed slice, so its numbers are not quoted on the slide. Full details and a second, unrelated scoping bug found in the dedup-aware spike path are in slide 8's speaker notes.

## Test plan

- [x] `uv run --extra serving --extra pipeline-core pytest` — 1982 passed, 19 skipped
- [x] `uv run ruff check .` — clean
- [x] Deck rebuilt (`build_deck.py`), converted to PDF, rendered to JPEG and reviewed slide by slide; slide 8 checked for text overflow after trimming item 2
- [x] Grepped every slide face for banned strings (`publication_ready`, `release_eligible`, the withdrawn numbers, em/en dashes) — no matches
- [x] `reference.pptx` confirmed not staged (gitignored, copied locally only)
- [x] 92 existing Sarvam-sample ticket ids confirmed present in the new 224-ticket draw, both in the manifest and on disk


ymohanty:
@codex review

PR #311, opened 2026-08-25
docs(presentations): split the routing-outcome slide into design and diagnosis

ymohanty:
Replaces slide 7, "The estimate did not survive the second year," with two slides seated between "Closing a case costs nothing" and "What we cannot measure." Deck goes from 10 slides to 11.

## Why

The single slide read as a feature that failed to ship. What happened is that a constrained policy-learning problem was specified, two duration models and a calibrated action classifier were fit, a cross-fitted AIPW estimator was built with censoring weights, and the overlap diagnostics refused the result on an untouched year. Slide 7 is now the design. Slide 8 is the diagnosis and the decision.

## Register

Codex's original brief was written for a government audience and barred notation, formulas and method on the slide face. The reader is an econometrician, so those rules were dropped. The estimator is named, and the diagnostics are on the slide rather than in the notes. No display math; the deck still has no equations.

## What changed on the face

Slide 7 clones the `A_NUM3` archetype, so the maroon 1/2/3 numerals carry the three stages with no authored shapes. It ends on the live/research firewall.

Slide 8 keeps the existing table idiom. Two substantive changes to the table:

- **Effective sample as a share of n** is new: 0.19 on 2024, 0.07 on 2025. It was absent from the deck and is the most diagnostic number available.
- The falsification row is relabelled and stays on the face. 5% error on 2024 against 118% on 2025, on the historical arm where truth is observed. That row is what localises the failure to positivity rather than to the year.

Censoring (9% / 34%) moves from the table into a caveat line, since it is a different threat from the four overlap rows. Interference is now named on the slide, because the design document concedes its no-interference assumption "in a queueing system is false."

Retitled to "Positivity failed before the second year did." The previous framing credited the action-rate constraint for a withdrawal that overlap caused.

## Numbers

None changed. Every figure still traces to `docs/experiments/routing-outcome-evidence-2026-08-19.json` and the 19 August refit under `outputs/experiments/routing_outcome/`. The withdrawn 11-to-23 day band stays withdrawn and appears only inside its do-not-cite warning. `docs/QUALITY_BENCHMARKS.md` status line is unchanged: no routing gain established.

## Also

- Slide 5's notes carried a full duplicate account of the outcome model. Trimmed to a pointer.
- Two stale slide-number references fixed: slide 6's notes pointed at "the next slide" for the failure, and slide 9's notes cited the outside option as "slide 6" when it was never slide 6.
- Banner comments and the README outline renumbered.

## Checks

- Builds: 11 slides, `SPEAKER_NOTES.md` regenerated with correct headings.
- Rendered via LibreOffice and inspected slides 7, 8 and 9. No overflow, no collision with the footer band, maroon conclusion last on both new slides.
- `ruff check` passes.
- No pytest: this directory has no importable package code.


ymohanty:
@codex review

PR #306, opened 2026-08-24
docs(presentations): add timing-and-quality technical briefing

ymohanty:
## Summary

- add a 12-slide timing-and-quality technical briefing for the demo
- keep claims tied to named evidence in speaker notes
- include the generator and generated PPTX

## Validation

- .venv/bin/python -m py_compile docs/presentations/2026-08-24-timing-quality/build_deck.py
- .venv/bin/ruff check docs/presentations/2026-08-24-timing-quality/build_deck.py
- git diff --check main...HEAD
- structural inspection: 12 slides, 11 slides with notes, 16:9 dimensions

## Draft blockers

- build_deck.py currently requires an uncommitted reference.pptx; the source reference contains citizen PII on portal-screenshot slides and cannot be added as-is
- the required presentation render runtime is unavailable in this environment, so the committed PPTX has not received the mandated visual QA pass
- handoff notes still say the descriptive mart is under recomputation even though PR #293 has now merged; source and generated deck must be updated together after a safe reference is available

This PR is intentionally draft and must not merge until the deck is self-contained or has a committed PII-safe template, regenerated, and visually verified.

PR #331, opened 2026-08-28
fix(evaluation): salvage the non-deck work from the retired deck PRs

ymohanty:
The presentation is finished, so #313 and #315 are being retired. Both carried real non-deck code that has nothing to do with slides, and this recovers it.

## What was extracted

Eleven commits, cherry-picked with `-x` so each keeps its original message and reasoning.

From **#315** (`fix/benchmark-stream-documents`) — all five commits were already deck-free:

- streaming staged document bytes instead of eager-loading them
- `StaleDocumentBytesError` and `_document_sample_file_hashes`, binding `sample_digest` to the bytes actually benchmarked
- OCR toolchain provenance in `benchmark_context`
- the base-CI skip and fingerprint-drift close
- the OCR probe test gated on the system binary, not just the Python wrapper

From **#313** (`fix/sarvam-schema-v2-and-benchmark-slide`) — six of eight commits:

- Extract schema v2: `district` dropped, category enumerated and required
- category label normalisation before scoring (the HTML-entity unescape)
- per-page disagreement recorded instead of page one winning
- one aggregator for both category paths
- the roadmap/DELIVERY record of the paid run

Its two deck-only commits (`6fdbba2`, `d181e30`) were skipped, and the deck hunks were stripped from the four mixed commits. **No file under `docs/presentations/` is touched by this branch.**

## Why the base is this branch and not main

Neither PR applies to `main`. Both build on the real-document input path that reaches main only through #326, and a cherry-pick of the first #315 commit onto main conflicts immediately. #326 already carries `docs/timing-quality-briefing` in full, so it is the one path these changes have.

## On #306

It needs no salvage at all: `docs/timing-quality-briefing` is entirely contained in #326 (`git merge-base --is-ancestor` confirms, 0 commits ahead). Every one of its commits already reaches main. It can be closed.

## Conflicts

Four, all additive — the two branches added code at the same points in `benchmark_pipeline.py`. Three were "keep both". The fourth was a real merge: `sample_digest` now comes from #315's `file_hashes=` form so it hashes the same bytes the run later verifies against, while keeping the `--slice`/manifest-conflict check and the coverage block from the review rounds on #326.

## Verification

`2013 passed, 19 skipped`, ruff clean — against 1980 on the base branch. The 33 additional tests are exactly the ones #313 and #315 shipped, passing against the merged code, which is what shows the strip dropped nothing and the conflict resolutions are right.

`Makefile` shows as differing from both retired branches; it is identical to `main`. That difference belongs to the deck branch and is already carried by #326.

### DIFF
--- a/docs/DELIVERY.md
+++ b/docs/DELIVERY.md
@@ -15,6 +15,18 @@ status: Internal
 > intake-time actionability and congestion remain unresolved, so it supports no
 > routing gain, deployable threshold, causal impact or recommendation. See
 > [QUALITY_BENCHMARKS.md](QUALITY_BENCHMARKS.md#routing--outcome-not-destination).
+>
+> **Sarvam spend resumed, 25 August 2026.** "No new paid calls" and "new spend
+> is paused" below describe the position up to 14 August and are superseded. A
+> 200-page stratified draw from Sambalpur/2024 ran on 25 August across both
+> endpoints, ₹300 at list price, 87 grievances. It supersedes the cached
+> aggregate and changes three statements below: a latency distribution now
+> exists (digitise median 6 s a page, extract 11 s, derived from the egress
+> audit log, 5 s polling resolution); divergence is 0.995 on 198 scored pages
+> with a character ratio of 1.56, not 1.3345 on 56; and a category grade now
+> exists but measures our own v1 schema, which offered no enum, rather than the
+> provider. Still absent, and still blocking any accuracy claim: a
+> hand-transcribed reference (#53). Schema v2 is now the pinned default.
 
 ## What is built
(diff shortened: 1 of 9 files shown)


ymohanty:
@codex review

ymohanty:
Picked up one more salvaged fix in c5011c5: the OCR-probe independence test stubbed poppler instead of requiring the binary (Codex P1 on #315).

`2013 passed, 19 skipped`, ruff clean. Verified with poppler on PATH and with it removed.

@codex review

ymohanty:
@codex review

ymohanty:
@codex review

PR #334, opened 2026-08-28
fix(rehearsal): guard the preflight argv against set -u on bash 3.2

ymohanty:
Salvage from `chore/demo-integration-freeze` before that branch is deleted. That branch was one commit, 231 behind `main`, with its remote already gone. Its content split three ways: the `outputs/benchmark/` `.gitignore` rules had already reached `main` independently, the freeze-tag rows are obsolete now that the 14 August demo has been delivered, and these two fixes had never landed anywhere.

## The bug

`scripts/demo_rehearsal.sh` runs under `set -euo pipefail`. It builds `PREFLIGHT_ARGS` as an empty array, appends `--strict` only when `REHEARSAL_STRICT=1`, then expanded it unguarded:

```bash
PREFLIGHT_ARGS=()
if [ "$REHEARSAL_STRICT" = "1" ]; then
  PREFLIGHT_ARGS+=(--strict)
fi
if ! uv run --extra demo janasunani-demo-preflight "${PREFLIGHT_ARGS[@]}"; then
```

On bash 3.2 — the system bash on the macOS demo laptop — expanding an empty array under `set -u` is an unbound variable:

```
$ /bin/bash -c 'set -u; A=(); printf "%s\n" "${A[@]}"'
/bin/bash: A[@]: unbound variable
```

So `make rehearsal` aborted before preflight ever ran, in the *common* case — non-strict. `PYTEST_FILES` two blocks earlier already had the length guard; this one did not.

## The fix

Build the command into `PREFLIGHT_CMD` behind a length check, mirroring the existing `PYTEST_FILES` guard.

## Tests

Both read the script off disk, so they track the file the operator actually runs rather than a copy that can drift.

- `test_demo_rehearsal_preflight_command_survives_set_u` executes the script's own `PREFLIGHT` block under `set -euo pipefail` and asserts the argv it builds, parametrised over both modes: `--strict` present when `REHEARSAL_STRICT=1`, absent otherwise.
- `test_demo_rehearsal_empty_arrays_are_length_guarded` states the portability invariant — every array the script initialises empty is read through a `"${#NAME[@]}"` guard. This one earns its place: CI runs bash 5, where the empty expansion is perfectly legal, so executing the block cannot catch a regression on its own.

Both fail against the previous script and pass against this one, verified by reverting the script and re-running.

## Also

Drops two stale references in `docs/DEMO.md` (lines 209 and 225) telling the reader that `make rehearsal` and `scripts/demo_rehearsal.sh` are still pending in PR #203. Both have been on `main` for weeks.

## Validation

- `uv run --extra serving --extra pipeline-core pytest` — 1965 passed, 19 skipped
- `bash -n scripts/demo_rehearsal.sh` — clean
- the shipped block executed under real `/bin/bash` 3.2.57, which is where the bug bites: now builds `uv run --extra demo janasunani-demo-preflight`

### DIFF
--- a/docs/DEMO.md
+++ b/docs/DEMO.md
@@ -206,8 +206,6 @@ The **13 August freeze** is gated by a single laptop command that proves every [
 ```bash
 make rehearsal
 # runs scripts/demo_rehearsal.sh — see docs/plans/2026-08-08-demo-integration-rehearsal.md Part 2
-# Note: the Make target and script land in PR #203 (chore/demo-rehearsal-script);
-# until that merges, run the individual Phase A–C checks from the plan directly.
 ```
 
 What it checks (four phases, fail-fast):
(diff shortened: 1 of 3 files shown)


ymohanty:
@codex review

PR #339, opened 2026-08-30
feat(evaluation): verified DSI corpus draws and a benchmark that names its sample

### LINKED ISSUE #335: Record dump: hold a descriptor on the validated directory so an ancestor swap cannot redirect it
Codex P1 on #326 (round fourteen), filed rather than fixed there.

## The gap

`_write_record_dump` in `janasunani/evaluation/sarvam_evaluate.py` now creates its temporary file with `tempfile.mkstemp(dir=destination.parent, ...)` and `os.replace`s it into position. That closed the symlink-planting hole from round thirteen — the basename is unguessable and created `O_CREAT|O_EXCL`.

It does not preserve the *containment decision* made at startup. `_checked_record_destination` validates that the path sits inside the governed `data/` tree, but `mkstemp(dir=destination.parent)` re-traverses the path later, during the paid run. A local process that replaces the validated parent directory with a symlink in between gets both the temporary file and the final dump created in the symlink target, outside `data/`.

The dump is unredacted per-page citizen text, so this is a PII-containment weakening, not just a correctness one. See `AGENTS.md` §"Data Access Restriction".

## The fix

Open the validated directory once at validation time and keep the descriptor:

- `dir_fd = os.open(validated_parent, os.O_RDONLY | os.O_DIRECTORY)`
- create the temporary with descriptor-relative, no-follow semantics (`os.open(..., dir_fd=dir_fd)`; `tempfile.mkstemp` has no `dir_fd`, so this needs an explicit `O_CREAT|O_EXCL|O_NOFOLLOW` loop over random names)
- `os.replace(tmp_name, destination.name, src_dir_fd=dir_fd, dst_dir_fd=dir_fd)`
- close the descriptor in a `finally`

Then no path component is resolved twice, and an ancestor swap after validation cannot redirect the write.

## Why it is filed rather than fixed

The threat model requires a local process with write access to the parent of a governed data directory, during a run — an attacker at that level plausibly has the corpus already. The fix is invasive on a code path that only runs under the optional `--save-records` debug flag, and #326 has taken fourteen review rounds, five of which were triggered by the previous round's own fix. This one is worth doing deliberately, on its own branch, with the descriptor lifetime reviewed properly — not as the fifteenth patch on a 4,900-line PR.

## Verification when picked up

`tests/test_sarvam_evaluate.py` already has `test_record_dump_temp_file_cannot_be_hijacked_by_a_symlink`. Extend it with a case that swaps the parent directory for a symlink between validation and write, and assert the target is untouched.

ymohanty:
Splits #326. This is the half where **every** one of its fourteen review rounds landed. The inert half is #338, which this is stacked on.

## Why this is its own pull request

Fourteen rounds, and five of them were caused by the fix from the round before — a marker guard that left the manifest mapping unchecked, a publication gate keyed off the wrong signal, an atomic write that introduced a symlink-planting path, a probe that no longer matched the write it validates. Each fix was correct and each changed an assumption elsewhere. That pattern is going to continue for at least another round or two, and the blast radius of it should be these six files rather than a 4,900-line PR with two others queued behind it.

## What is here

**`janasunani/evaluation/dsi_sample.py`** — nested stratified draws over the DSI reference corpus, with the manifest verification that makes a draw *verified* rather than *whatever happens to be on disk*. It rejects duplicate `s3_key`s, blank identities, a ticket that contradicts its key, a manifest with no rows, and — via `expected_documents` — a manifest and a corpus that agree with each other while both being a strict subset of the pinned population. That last one is the failure mode the whole module exists to prevent, and the set comparison alone cannot see it because it is symmetric.

**`scripts/benchmark_pipeline.py`** — real staged documents as a benchmark input, with a publication gate that requires an *explicit* sample verdict. Silence is a refusal: a run declares `synthetic` or `staged-documents` provenance, and a staged run must name a slice that is not `"unspecified"`, carry a digest, and account for every manifest row. Rows are validated against their own `s3_key`, prefix against `ticket` and basename against `file` — never the two against each other, because a hierarchical ticket stages under `Path(key).name` and a naive check would reject exactly the ~2% of complaints the manifest exists to protect.

**`janasunani/evaluation/sarvam_evaluate.py`** — the record dump is written to an unguessable, exclusively created temporary and replaced atomically, so a failed or interrupted write cannot destroy paid-run evidence and a planted symlink cannot divert unredacted citizen text out of the governed tree.

## Dependency on #338

One edge: `tests/test_dsi_sample.py` pins the module docstring's corpus count against `REGISTRY["dsi_large"]`. That guard exists because the docstring had drifted to the known-short Box copy's 69,844, which would have made a caller draw from a population 185 documents short. Retarget to `main` once #338 lands.

## Known, filed rather than fixed

#335 — hold a descriptor on the validated directory so an ancestor swap mid-run cannot redirect the record dump. Real, deferred deliberately; reasoning is in the issue and in the #326 thread.

## Validation

- `uv run --extra serving --extra pipeline-core pytest` — **2102 passed, 19 skipped**

Every file is byte-identical to its state on #326 at `08aad9e`, so the review history there carries over.

### DIFF
--- a/janasunani/evaluation/dsi_sample.py
+++ b/janasunani/evaluation/dsi_sample.py
@@ -1,0 +1,594 @@
+"""Nested stratified draws over the DSI clinic reference corpus.
+
+The DSI clinic's ``large_sample`` is the corpus their reported numbers were
+measured on: **70,029 documents across 69,977 tickets**, drawn from 100,000
+complaints at seed 1337 (see :mod:`janasunani.samples`, which pins it).
+Reproducing anything of theirs means working from it rather than from a slice
+of our own choosing.
+
+69,844 is *not* this number. It is the count in the known-short Box copy,
+which is missing 189 documents that S3 has, and it appears in this file only as
+the failure that :func:`load_corpus`'s manifest comparison exists to catch.
+Passing it as the largest tier's budget would make ``_draw`` sample 185
+documents out of the verified population instead of returning it whole, and
+every nested tier would then descend from an incomplete corpus --- the same
+wrong-population failure, arrived at through the budget rather than the disk.
+
+It is also 60 GB, and a single pipeline pass over all of it is days of compute.
+So the work is tiered, and the tiers must nest:
+
+    latency  (few hundred)  subset of
+    quality  (few thousand) subset of
+    corpus   (len(corpus), never a copied constant)
+
+**Nesting is by construction, never by re-seeding.** A seeded re-draw at a
+smaller budget is not a subset of the larger one: ``allocate`` apportions a
+per-category floor first and then the remainder by largest remainder, so
+changing the budget reshuffles which categories get headroom and the two draws
+share only what they happen to share. This module therefore draws the largest
+tier from the corpus, then each smaller tier *from the tier above it*. That is
+the only way the latency numbers and the quality numbers describe the same
+documents, which is the whole point of doing it this way.
+
+The stratifier is ``sarvam_sample_builder.allocate``, reused rather than
+reimplemented. Its docstring records three failures it already survived; a
+second copy here would get to rediscover them.
+
+Reads only the paths it is given. Emits manifests of ticket ids, categories and
+filenames -- never document text.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import random
+import re
+import unicodedata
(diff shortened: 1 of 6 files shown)


ymohanty:
@codex review

Issue #158, opened 2026-08-07
dedup grouping is quadratic in bucket size: 378M pairs on one district-year


ymohanty:
Found running the index on Sambalpur 2024. Memory is fixed (#152, #153); **time is not.**

## The measurement

From the 55,544 persisted signatures:

```
non-singleton buckets   46,191
largest bucket           6,797 members
total pairs to walk    378,302,087
top bucket sizes    6797, 6481, 6119, 5561, 5537, 5536, 5519, 4965
```

The grouping stage has been running **55 minutes of CPU** on one district-year and has not finished.

## Why it is slow even with the union-find short-circuit

`itertools.combinations` still *generates* every pair. A 6,797-member bucket is 23 million tuples produced and discarded even when `find(a) == find(b)` skips all of them. The short-circuit avoids the expensive Jaccard verification; it does not avoid the enumeration.

Second cost: one database round trip per bucket, and membership overlaps across 22 bands, so the same ticket's text is fetched roughly 22 times. 46,191 round trips.

## Why the buckets are so large

This is a campaign-heavy corpus. Near-identical filings share band hashes by design — that is LSH working. The failure is that the grouping algorithm assumes buckets are small.

## Options

1. **Representative comparison above a size cap.** Per bucket, keep one representative per component discovered so far; compare each new member against representatives only. O(members x components) instead of O(members squared), and components collapse fast in exactly the campaign case that produces big buckets. **Costs recall**: a member that would verify against a non-representative but not the representative is missed. That is the same trade #101 argued against, so the cap should be high enough that small buckets stay exhaustive.
2. **Cache text per block rather than per bucket.** Removes the 22x refetch without changing semantics. Helps constant factors, not the quadratic.
3. **Raise the similarity threshold or narrow the bands** so buckets are smaller. Changes what counts as a duplicate, which is a measurement decision rather than a performance one.

**Option 1 plus option 2** is the combination I would take, with the cap stated in the artifact and in #72's write-up so the recall trade is visible rather than buried.

## Why this matters beyond one run

#71 requires the index to be re-runnable, and #136 adds `--refresh-stale`, which rebuilds everything. A stage that takes hours per district-year makes both theoretical. The full corpus is 24x this slice.

## Related

- #152, #153 — the two memory fixes; this is the same stage, different resource
- #101 — the all-pairs invariant a size cap would partially trade away
- #71 — the backfill this blocks
- #72 — consumes the groups

PR #340, opened 2026-08-30
Bound corpus-wide grouping memory: per-block verification and a bounded identity path

### LINKED ISSUE #336: Corpus-wide grouping needs ~59 GB: two unbounded caches in the verification pass
`--all` grouping cannot complete on the box. `d11a8aa` removed the 16N band map, which was a real 4 GiB consumer but not the binding one. The binding consumers are two unbounded caches in the verification phase of `_group_duplicates`.

## Measured, not estimated

Sambalpur/2024 on `fe0d105`, the largest district-year in the corpus (55,544 of 1,361,842 signatures). Signatures were already indexed, so this timed grouping alone.

| | |
|---|---|
| Wall time | 44m 47s |
| RSS before grouping | 788 MB |
| Peak RSS | 3,169 MB |
| Grouping-attributable | 2,381 MB |
| Duplicate groups | 10,874 |
| Comparison pairs | 16,638,274 |
| Large buckets | 322 |

A `py-spy` dump taken mid-run put the process in `jaccard_similarity` / `_verify_bucket` / `_group_duplicates:1476`, past `_candidate_buckets`. The memory that was still climbing at that point was not band accumulation.

## The two caches

Both are built once for the whole scope and live for the entire verification pass:

- `text_by_ticket` (`dedup_index.py:1462`) holds the redacted text of every candidate ticket in scope. `candidate_tickets` is assembled across all blocks and all identity buckets before a single bucket is verified.
- `shingle_cache: dict[str, set[str]]` (`dedup_index.py:1468`) holds a shingle set per ticket and is threaded through every `_verify_bucket` call. Nothing evicts from it.

Both scale with candidate tickets rather than with bucket size. Shingle sets are the expensive half: a set of k-gram strings per document, several KB each.

## Projection to `--all`

`build_dedup_index` passes `(None, None)` for a corpus run, so `_group_duplicates` runs once over all 1.36M rows rather than per slice. The corpus is 24.5x this slice and both caches scale close to linearly in candidate tickets.

- Memory: about **59 GB**, against 15.7 GB on the box.
- Time: at least **18 hours**, and that is a floor. Identity buckets are deliberately unblocked, so they grow faster than linearly once the scope spans every district.

This matches the observed failure. The 2026-08-28 corpus run reached 15 GB and was still climbing when the kernel OOM killer took it at 20:53:28 UTC (`exit=137`). It ran pre-fix code, but the fix would not have saved it.

## Proposed fix

Buckets are already grouped by block in `band_buckets_by_block`, and `block_key` is `district:script:window`, so a ticket almost never spans blocks. Load text per block and clear `shingle_cache` between blocks. That bounds the peak at one block instead of the whole corpus.

Identity buckets genuinely do span blocks, by design (see the module docstring). They are a much smaller set and can keep the current behaviour.

This trades away part of the #158 optimisation of fetching each block's candidate text once, so it costs additional database round trips. It does not address the 18 hours, which is a separate question.

## Notes

- Indexing is unaffected and complete. All 1,361,842 signatures are in `dedup_signatures` and survived the OOM, because batches commit as they go.
- Per-slice runs are fine at current sizes. The largest slice peaks at 3.2 GB with 11.8 GB free.
- Do not restart `--all` on the box before this lands.

ymohanty:
Corpus-wide grouping needed about 59 GB and could not run. It now peaks at 692 MB on the largest district-year, with bit-identical results.

## Measured on the box, same slice, same sampler

Sambalpur/2024, the largest district-year (55,544 of 1,361,842 signatures). Signatures were already indexed, so each run timed grouping alone.

| | `fe0d105` before | `09bed95` per-block | `c667604` + identity |
|---|---|---|---|
| **peak RSS** | 3,169 MB | 2,405 MB | **692 MB** |
| min free | 11,814 MB | 12,587 MB | 14,291 MB |
| duplicate groups | 10,874 | 10,874 | 10,874 |
| comparison pairs | 16,638,274 | 16,638,274 | 16,638,274 |
| large buckets | 322 | 322 | 322 |
| wall time | 44m 47s | 41m 41s | 41m 21s |

**The answer never moves.** Groups, comparisons and large-bucket counts are identical across all three, which is the property that matters: this is a memory change, not an algorithm change.

The measured run was `179d325`; `c667604` is that tree plus an import reorder and one blank line, amended before review.

## Two commits

**`09bed95` — verify one block at a time.** `_band_buckets` keys every bucket on `(block_key, band_hash)`, so a band bucket can never span two blocks. Finishing a block before starting the next is the same set of buckets in a different order, and union-find connectivity does not depend on arrival order. Candidate text and the shingle cache now live for one block instead of for the run, which is what `_verify_bucket`'s docstring already claimed ("see `_group_duplicates`'s per-block cache") — the cache was created outside the block loop and never was per-block.

This bounds the peak because a block does not grow with the corpus. `block_key` is district:script:window, and over all 1,361,842 signatures there are 3,227 blocks, mean 422, p99 3,268, largest 8,691.

`_candidate_buckets` becomes a generator pair and stays as the whole-scope statement the bucket tests assert against. Production no longer calls it, and a guard asserts that.

**`c667604` — bound the identity path.** The per-block split fixed the band phase and left identity verification as the ceiling: the band phase peaked at 779 MB, and a `py-spy` dump caught the rest inside `_verify_identity_batch` at 2,053 MB.

Identity buckets are far more uneven than "one citizen's filings" suggests:

| | buckets | members | largest |
|---|---|---|---|
| slice | 5,840 | 52,824 | 27,641 |
| corpus | 13,428 | 1,355,855 | 201,965 |

So `_verify_bucket` no longer memoises non-anchors on the large-bucket path — a non-anchor is read once, against every anchor, and caching it grows the cache by the size of the bucket (~4 GiB for the 201,965-member one). An entry another bucket already paid for is still reused; this only stops new ones being added. The exhaustive path below `cap` is unchanged, so #158's reuse survives where it pays. And identity batching became a member budget rather than a bucket count.

## Projection for `--all`

banded rows ~1.05 GB, union-find ~0.2 GB, largest block ~0.2 GB, identity batch ~0.5 GB, interpreter ~0.4 GB. **About 2.2 GB, flat in corpus size**, against 15.7 GB on the box.

Runtime is untouched. The corpus is still an ~18 hour job. Parallelising across now-independent blocks is a separate question and not attempted here.

## Costs, both accepted

One round trip per block instead of one per run, and re-shingling a ticket that appears in both a band and an identity bucket.

## Base branch

Cut from `main`, not from `docs/timing-quality-briefing` or the current working branch. Those are 22 commits behind `main` and do not contain `d11a8aa`, so their `dedup_index.py` is the pre-fix version and the diff would not have been reviewable against the real code.

## Verification

- `uv run --extra serving --extra pipeline-core pytest` — 2,120 passed, 8 skipped. Not against production Postgres.
- `uv run ruff check .` — clean.
- Two full box runs on the real slice, above.
- No `dvc pull`; no gold-metric gate is touched by this change.

## Not fixed here

One hashed mobile accounts for 201,965 signatures, and non-singleton identity buckets cover 1,355,855 of 1,361,842. A citizen did not file 201,965 grievances — that is a placeholder number being treated as an identity, and the identity path is doing large amounts of work on pairs that are not same-citizen. This change makes that survivable, not correct. Filed separately.

Closes #336

### DIFF
--- a/janasunani/pipeline/dedup_index.py
+++ b/janasunani/pipeline/dedup_index.py
@@ -180,6 +180,7 @@ from dataclasses import dataclass
 from datetime import date, datetime, timezone
 import hashlib
 from itertools import combinations
+from collections.abc import Iterator
 from typing import Optional
 
 from loguru import logger
@@ -274,6 +275,15 @@ LARGE_BUCKET_ANCHOR_COUNT = 32
 # Stable provenance marker for the bounded grouping policy. Increment this
 # when the algorithm changes even if the cap/anchor defaults do not.
 GROUPING_ALGORITHM = "fixed-anchor-v1"
+
+# Identity buckets carry no block key, so they cannot be scoped to a block
+# the way band buckets are (#336). They are bounded by batching instead: each
+# bucket is one citizen's filings, so this many of them is a small working
+# set even though their members can come from anywhere in the scope. Not a
+# tuning knob -- it trades round trips against peak memory, and only the
+# memory side has a hard limit.
+IDENTITY_BUCKET_BATCH = 2_000
+
 
 # Odia Unicode block. Presence, not majority: any real Odia content in a
 # filing is enough to route it to the Odia-script partition rather than the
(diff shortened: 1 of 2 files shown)


ymohanty:
@codex review

ymohanty:
Follow-up finding while validating this, recorded in #341: `petitioner_mobile` is masked in OLTP, and 0 of 1,359,804 identity-keyed signatures have a phone-shaped value. The 201,965-member bucket is the export sentinel `~::~`. Email linkage is unaffected and healthy.

Nothing here changes this PR — it is a memory fix and stands on its own. But it does mean the identity work this PR makes survivable is mostly work on masked values, and fixing #341 would cut corpus runtime substantially on top of the memory win.

ymohanty:
@codex review

ymohanty:
Pushed an empty commit `6ba517c` to move the head. No content change.

The gate creates its `codex-review` check through the API, which attaches it to the app`s *first* check suite for the commit. This branch was open across three days of workflow runs, so `c667604` accumulated twelve check suites in three batches and the verdict stayed anchored in the oldest one:

| batch | holds `codex-review`? |
|---|---|
| 30 Aug 16:39 | yes |
| 31 Aug 10:21 | no |
| 1 Sep 07:02 | no |

The ruleset looks in the current batch, does not find it, and reports the check as `expected` — with every required check green and Codex having passed. Re-running the gate did not help, because a fresh check run lands in the same stale suite, and `no-push-only-pr` has no bypass actors so an admin merge is refused too.

Moving the head is the way through: the new commit gets one batch of suites and the verdict lands inside it, which is the state every recently-merged PR was in. The underlying gate defect is #345.

@codex review

